Issue #430, opened 2026-05-20
CLI: Route all output through CLIOutput — eliminate raw print() calls


malpern:
## Problem

3 commands bypass the `CLIOutput` system with raw `print()`:
- `ConfigPathCommand.swift` — `print(path)`
- `ConfigShowCommand.swift` — `print(config)`
- `HelpExamplesCommand.swift` — multiple `print()` calls

This means `--json` doesn't wrap their output in the `apiVersion` envelope, `--quiet` doesn't suppress them, and agents get inconsistent response shapes.

## Fix

Route through `CLIOutput.write()`:
- `config path --json` → `{"path": "/Users/.../.config/keypath/keypath.kbd"}`
- `config show --json` → `{"content": "...", "path": "..."}`
- `help-topics examples --json` → structured array of example objects

## Effort: Low (~30 minutes)

malpern:
Fixed in PR #436. `config path` and `config show` now route through CLIOutput with GlobalOptions, supporting `--json`.

Issue #427, opened 2026-05-20
CLI: Wrap errors in apiVersion envelope and standardize error handling


malpern:
### Error envelope missing
JSON errors go to stderr without the `apiVersion` wrapper. An agent checking for `{"apiVersion": 1, ...}` gets nothing on failure — errors are bare `{"code": ..., "message": ...}` objects.

**Fix:** Errors should use the same envelope: `{"apiVersion": 1, "error": {"code": ..., "message": ..., "hint": ...}}`

### Inconsistent error types leak to users
Some commands leak raw `ValidationError` (from ArgumentParser) or domain exceptions like `AmbiguousCollectionMatch` that bypass `CLIOutput.writeError()`. An agent parsing stderr gets a Swift error dump instead of structured JSON.

**Fix:** Add a catch-all error handler (protocol extension or root command wrapper) that wraps any `Error` in `CLIError.validation()` as a last resort.

### Breaking vs. additive changes undocumented
Adding `changeset` to `CLIApplyResult` is additive (safe). Changing `action` values in `CLIEnsureResult` would be breaking. This contract should be written down.

## Effort: Medium

malpern:
Fixed in PR #436. JSON errors now wrapped in `{"apiVersion": 1, "error": {...}}` envelope, matching the success path.

Issue #426, opened 2026-05-20
CLI: Fix MainActor isolation — CLIFacade created off-MainActor in 7 commands


malpern:
## Problem

7 CLI commands create `CLIFacade()` directly without `MainActor.run`, but the facade has 9 `@MainActor` methods (`validateConfig()`, `runStatus()`, `runInstall()`, etc.). This works today because Swift's runtime is lenient, but **SE-0414 strict concurrency checking** will enforce isolation boundaries and these paths will crash.

**Commands creating CLIFacade off-MainActor:**
- `ConfigCheckCommand.swift`
- `ServiceLogsCommand.swift`
- `ServiceStatusCommand.swift`
- `SystemInspectCommand.swift`
- `SystemInstallCommand.swift`
- `SystemRepairCommand.swift`
- `SystemUninstallCommand.swift`

## Fix Options

1. **Mark CLIFacade as `@MainActor`** — honest about what it needs, all callers use `await MainActor.run { CLIFacade() }`
2. **Remove `@MainActor` from facade methods** — push isolation to the specific internal callsites that need it (ConfigurationService, InstallerEngine)
3. **Make CLIFacade `Sendable`** with internal synchronization

Option 2 is preferred — the facade itself doesn't need MainActor, only some of its dependencies do.

## Effort: Medium

malpern:
Fixed in PR #436. All 7 commands now use `await MainActor.run { CLIFacade() }` consistently.

PR #436, opened 2026-05-20
fix: MainActor isolation, error envelope, output consistency (#426 #427 #430)

malpern:
## Summary

Addresses three issues from the CLI architecture audit:

- **#426 (P0) MainActor isolation**: 7 commands created `CLIFacade()` without `MainActor.run`, calling `@MainActor` methods from a non-isolated context. This will crash under SE-0414 strict concurrency. Fixed by standardizing all commands to use `await MainActor.run { CLIFacade() }`.

- **#427 (P1) Error envelope**: JSON errors went to stderr as bare `CLIError` objects without the `apiVersion` wrapper. Agents checking for `{"apiVersion": 1, ...}` got nothing on failure. Now errors use `{"apiVersion": 1, "error": {...}}` — matching the success envelope.

- **#430 (P3) Output consistency**: `config path` and `config show` used raw `print()`, bypassing `CLIOutput`. Now they route through `CLIOutput.write()` with `@OptionGroup var globals`, so `--json` returns `{"path": "..."}` and `{"content": "..."}` respectively.

## Test plan
- [x] `swift build` — clean
- [x] `swift test` — 530 tests pass
- [x] Verified all 46 command files now use `await MainActor.run { CLIFacade() }` consistently
- [x] No bare `CLIFacade()` calls remain in command files

### DIFF
--- a/Sources/KeyPathAppKit/CLI/CLIFacade.swift
+++ b/Sources/KeyPathAppKit/CLI/CLIFacade.swift
@@ -626,11 +626,29 @@ public struct CLIFacade: Sendable {
     public func validateConfig() async -> CLIValidationResult {
         let service = ConfigurationService()
         let config = await service.current()
+        let path = service.configurationPath
+        let collections = await RuleCollectionStore.shared.loadCollections()
+        let customRules = await CustomRulesStore.shared.loadRules()
+
         if config.content.isEmpty {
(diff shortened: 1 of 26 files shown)


malpern:
Superseded by #439 (CLI quality batch) which included all these changes and was already merged.

PR #518, opened 2026-05-22
fix: suppress orange flash on overlay during layer switch

malpern:
## Summary
- Clear stale `layerKeyMap`/`remapOutputMap` on layer switch so keys from the previous layer don't briefly flash with collection colors (F1-F12 and space bar flashing orange when holding F to enter nav layer with ben pack)
- Gate layer-mode collection coloring on `collectionId != nil` so pass-through keys without a collection fall to dark gray instead of default orange
- Pressed keys with a zone color keep that color instead of turning blue

## Test plan
- [x] 532 tests pass
- [ ] Enable ben pack, hold F — verify no orange flash on F1-F12 or space bar
- [ ] Verify nav layer keys still show correct collection colors
- [ ] Verify activator key (F) stays its zone color when pressed


malpern:
Closing — this fix was already merged via #515.

PR #645, opened 2026-05-30
feat(lifecycle): wait for kanata exit before start to fix stop→start race (#625 part-1)

### LINKED ISSUE #625: kanata restart can leave input degraded: keyboard-grab release race + no grab-failure detection/retry
## Problem
Toggling the service OFF→ON (and quick-deploy restarts) can start a **new kanata before the previous instance has released its exclusive keyboard grab**. The new instance fails to grab ("Another process is already grabbing your keyboard exclusively"), comes up **degraded** (TCP server + VHID output ready, but no input capture), and **never retries** — remapping is silently dead until a forced clean restart.

Observed this session: an OFF/ON toggle produced a kanata that couldn't grab; `sudo launchctl kickstart -k system/com.keypath.kanata` (clean kill+restart with the predecessor gone) was needed to recover.

## Root cause
The stop→start sequence doesn't guarantee the old kanata's process is gone (grab released) before the new one attempts to grab. There's no post-start verification that the grab succeeded, and no retry on grab failure.

## Proposed fix
1. In the stop path (`ServiceLifecycleCoordinator`), wait until the kanata process is actually gone (grab released) before starting a new instance.
2. After start, **verify the grab succeeded** (tie into the input-liveness signal from #624) and **retry** (or surface a clear error) on failure instead of leaving a degraded instance.
3. Detect the "failed to grab" stderr phrases and auto-recover.

## Severity
HIGH — a routine OFF/ON or deploy can silently break remapping. Pairs with #624 (which would at least make the degraded state visible).

malpern:
## Summary

Fixes the **stop→start race** that lets kanata enter the silent **degraded state** (process up + TCP responding, but it never seized the keyboard, so remapping is dead). This is **part-1** of #625 — the race fix at the source. **Part-2** (auto-recovery safety net, #644) is already merged; this completes the issue.

**Root cause:** `restartKanata()` does stop-then-start. The old kanata's termination is asynchronous — `SMAppService.unregister()` returns before the OS finishes tearing the process down, and the old `killOrphanedKanataProcesses()` only slept a fixed 500ms. So when `startKanata()` → `register()` runs, the old kanata can still be alive holding the **exclusive keyboard grab** and **TCP port 37001**. The new kanata then can't grab → degraded.

**Fix:** Replace the fixed-sleep best-effort with a bounded, deterministic `waitForKanataExitBeforeStart()` that runs right before `register()`:

- `pgrep` orphans → `SIGTERM` → poll liveness (`kill(pid,0)`, 100ms interval) until gone → `SIGKILL` survivors past a 2s grace → brief confirm-gone poll.
- **Only if** something was killed, poll `TCPProbe` until port 37001 is released (2s budget). Clean machine (no orphans) = **zero added latency**.
- **On timeout: warn and proceed** rather than failing the start — leaving the user with no remapping would be worse, and part-2's grab auto-recovery is the backstop for any residual case.

**Why it's safe w.r.t. the part-2 machinery:**
- The part-2 gate (`stopGraceUntil` / `intentionalStopDepth`) is **untouched**.
- The wait sits entirely between `AppContextService.stop()` and `.start()`, so the TCP grab-event listener is down throughout — **no grab events are processed during the wait**, so the longer wait creates no new spurious-recovery window.

## Testing

- 7 new deterministic tests in `ServiceLifecycleCoordinatorTests` drive the polling logic via DEBUG seams (pgrep/liveness/signal/TCP/sleep) — **no real subprocess, signal, port, or wall-clock sleep**. Cover: no-orphans fast path, gone-immediately, linger-then-exit, never-exits→SIGKILL→proceed, port-busy-then-frees, port-never-frees→proceed, safe-to-call-repeatedly.
- `pgrepMatches` additionally hard-refuses real `pgrep` under `TestEnvironment.isRunningTests` (closes a latent parallel-run deadlock).
- Full `swift test` suite green; `SKIP_NOTARIZE=1 ./build.sh` produces a signed app.

**Manual verification (post-merge):** exercise a restart (toggle service / change a rule), confirm via `~/Library/Logs/KeyPath/keypath-debug.log` that the wait-for-exit log lines appear, the new kanata grabs the keyboard (no degraded state), and no spurious auto-recovery fires during a normal restart.

## Review

Ran `/code-review high` (7 finder angles) against the diff and addressed the actionable findings: removed two vacuous tests + replaced with an honest direct test; hardened the test-pgrep seam; dropped a wasted trailing sleep on the port-timeout path. Pre-existing items (pgrep `-f` breadth, PID-reuse window, overlapping-restart concurrency) noted as out of scope. The mandatory `/thermo-nuclear-swift-review` cloud gate was not run by the agent (user-triggered) — worth running on this PR.

Fixes #625

### DIFF
--- a/Sources/KeyPathAppKit/Managers/ServiceLifecycleCoordinator.swift
+++ b/Sources/KeyPathAppKit/Managers/ServiceLifecycleCoordinator.swift
@@ -54,7 +54,7 @@ final class ServiceLifecycleCoordinator {
     /// Timing for the stop→start wait-for-exit (#625 part-1). Bounds are expressed in
     /// milliseconds so poll counts derive trivially and stay deterministic when the
     /// sleep seam is a no-op (a wall-clock deadline would never advance under that seam).
-    private enum WaitForExitTiming {
+    enum WaitForExitTiming {
         static let pollIntervalMs = 100
         static let processExitGraceMs = 2000 // SIGTERM → SIGKILL window
         static let portReleaseTimeoutMs = 2000 // after processes confirmed dead
@@ -344,6 +344,10 @@ final class ServiceLifecycleCoordinator {
     /// Bounded: on timeout we log a warning and proceed anyway rather than failing the
     /// start — leaving the user with no remapping would be worse, and the #625 grab
     /// auto-recovery is the backstop for any residual degraded case.
+    ///
+    /// `internal` (not `private`) so tests can drive it directly: `startKanata`'s VHID
+    /// gates short-circuit in the test environment, so the seam-injected polling logic
+    /// is exercised by calling this method rather than through `startKanata`.
     func waitForKanataExitBeforeStart() async {
         let processNames = ["kanata-launcher", "kanata"]
         var killedAny = false
@@ -374,9 +378,9 @@ final class ServiceLifecycleCoordinator {
 
         // Only wait on the TCP port if we actually killed something. A clean machine has
         // nothing of ours holding the port, so the common first-start path adds no latency.
-        if killedAny {
-            await waitForPortReleased()
-        }
+        guard killedAny else { return }
+        await waitForPortReleased()
+        AppLogger.shared.log("✅ [Service] wait-for-exit complete — old kanata gone, proceeding with start")
     }
 
     /// Poll the given PIDs until none are alive or the budget elapses.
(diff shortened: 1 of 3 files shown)


malpern:
Addressed the review feedback in `c609c787`:

- **EPERM liveness (Codex P2)** — `isProcessAlive` now treats `kill(pid,0)==EPERM` as **alive** and only `ESRCH` as gone. This was a real bug: kanata is a root LaunchDaemon and the app is unprivileged, so the old `== 0` check reported the live root process as dead and let the wait return immediately — exactly the process the fix exists to wait out. Fixed.
- **Off-by-one extra sleep (#1)** — `waitForProcessesGone` now skips the sleep after the final poll (mirrors `waitForPortReleased`), so the budget is ~`budgetMs` rather than `budgetMs + 1 interval`; final return no longer re-probes.
- **Success-path observability (#4)** — added `✅ [Service] wait-for-exit complete …` so a clean run is visible in `keypath-debug.log`.
- **`internal` intent (#3)** — documented why `waitForKanataExitBeforeStart` is `internal` (test access; `startKanata`'s VHID gates short-circuit in tests).
- **Hardcoded `20` (#6)** — exposed `WaitForExitTiming`; the port-timeout test now derives its bound from `pollCount(forMs: portReleaseTimeoutMs)`.
- **Both-names coverage (#5)** — added `testWaitForExit_bothProcessNamesHaveOrphans`.

Not changed: the 150ms-probe/100ms-sleep note (#2) — the composition makes the cadence *slower* than the interval, not faster, so it's benign. The `/thermo-nuclear-swift-review` gate is user-triggered (the agent can't launch it); flagging for a maintainer to run before merge.

Issue #463, opened 2026-05-21
fix: detect leader key vs collection activator conflicts


malpern:
## Problem

The leader key preference generates a `layer_{layer}_{key}` alias independently of collections (in `buildCollectionBlocks()` lines 63–117). If a collection also has a momentary activator targeting the same key and layer, both would produce the same alias name.

## Current protection

- `seenActivators` set in `buildCollectionBlocks()` prevents the second activator from being emitted — but this is silent dedup, not detect-and-explain
- `detectConflicts()` doesn't check leader key preference inputs at all

## Proposed fix

Include the leader key preference's input key in the conflict detection pass so it's treated like any other key claim. If a collection's momentary activator conflicts, surface it to the user.

## Risk

Low — only one leader key exists architecturally, but not validated. Could become an issue if the UI allows per-collection leader keys in the future.

## Related

- #458, #459 — conflict detection infrastructure

PR #658, opened 2026-05-31
fix: detect momentary-activator conflicts instead of silent dedup (#466)

### LINKED ISSUE #466: fix: make momentary activator dedup detect-and-explain instead of silent
## Problem

When two collections have momentary activators targeting the same key, buildCollectionBlocks() uses a seenActivators set to silently drop the second one. This is the same silent resolution pattern we eliminated for HRM/LayerToggle conflicts (#458) and chord group conflicts (#459).

## Current behavior

The second collection's activator is silently dropped — no error, no user notification. The user sees one collection's layer activation working and the other not, with no explanation.

## Proposed fix

Move the activator conflict check into detectConflicts() so it follows the detect-and-explain pattern: surface which collections have conflicting activators and let the user decide which to keep.

## Risk

Low — most collections have unique activator keys, but possible if two navigation-style collections are enabled simultaneously.

## Related

- #458 — established detect-and-explain pattern for cross-collection conflicts
- #459 — same pattern for within-collection chord conflicts

malpern:
## Summary

Closes the silent-dedup gap from issue #466 (part of the ADR-039 conflict-detection cluster).

When two enabled collections claim the **same physical activator key** but route it to **different layers**, `buildCollectionBlocks`'s `seenActivators` set silently dropped one. The user saw one collection's layer activation work and the other vanish, with no explanation — a violation of [ADR-039](docs/adr/adr-039-key-conflict-resolution-principles.md) §1 ("never silently discard a user's intent").

This moves the check into `RuleCollectionDeduplicator.detectConflicts()`, so the conflict surfaces through the existing `MappingConflictInfo → KeyPathError.mappingConflicts` thrown-error gate **before config generation** (ADR-039 §2, "detect at the earliest moment"), consistent with how cross-collection key overlaps and chord-group conflicts already surface.

## Behavior

- **Conflict** — two enabled collections whose normalized activator keys match but target **different** layers. Surfaced with the collection names and which layer each activates.
- **Not a conflict** — same key + **same** target layer. These are redundant (both want the same layer activated the same way); dedup harmlessly keeps one. This matches the existing toggle-time `conflictInfo` semantics, so detection doesn't over-fire.
- `hyper` activators and disabled collections are skipped, consistent with generation.

## Scope

Collection-vs-collection only. **Leader-key-vs-activator** collisions (the system leader key colliding with a collection activator) are deferred to **#463**, which threads `leaderKeyPreference` into `detectConflicts()`.

## Tests

Added to `RuleCollectionDeduplicatorTests`:
- conflict surfaced when two activators target different layers
- no conflict when they target the same layer (redundant)
- no conflict for distinct keys
- disabled-collection activators ignored

Full suite green (`swift test`), SwiftFormat 0.61.1 clean on changed files.

Fixes #466

### DIFF
--- a/Sources/KeyPathAppKit/Models/RuleCollectionDeduplicator.swift
+++ b/Sources/KeyPathAppKit/Models/RuleCollectionDeduplicator.swift
@@ -58,34 +58,46 @@ enum RuleCollectionDeduplicator {
         }
 
         // Momentary activator conflicts (#466): two collections claim the same
-        // physical activator key but route it to different layers. The silent
-        // `seenActivators` dedup in buildCollectionBlocks would drop one without
-        // notice — surface it via the same detect-and-explain path instead.
-        // Identical activators (same key + same target layer) are redundant, not
-        // conflicting, and are intentionally skipped (matching the toggle-time
-        // conflictInfo semantics).
+        // physical activator key from the same source layer but route it to
+        // different target layers. The silent `seenActivators` dedup in
+        // buildCollectionBlocks would drop one without notice — surface it via the
+        // same detect-and-explain path instead.
+        //
+        // The claim slot is (sourceLayer, input): a key can only carry one binding
+        // per layer. The same key activating different targets from *different*
+        // source layers is a valid chained-layer setup (e.g. `f` → arrows from base
+        // and `f` → function from nav), not a conflict. Identical activators (same
+        // slot + same target) are redundant, not conflicting. Both are skipped so
+        // the detection does not over-fire (matching the toggle-time conflictInfo).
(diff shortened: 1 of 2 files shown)


malpern:
Good catch — fixed in 868aabc1. Activator claims are now keyed by `(sourceLayer, input)`, so the same key activating different targets from different source layers (e.g. `f` → arrows from base, `f` → function from nav) is correctly treated as a valid chained-layer setup rather than a conflict. The surfaced conflict also now reports the actual source layer instead of a hardcoded "Base". Added a regression test (`testNoActivatorConflictWhenSameKeyDifferentSourceLayers`).

PR #659, opened 2026-05-31
fix: detect leader-key vs collection-activator conflicts (#463)

### LINKED ISSUE #463: fix: detect leader key vs collection activator conflicts

malpern:
## Summary

Part of the ADR-039 conflict-detection cluster; builds directly on #466.

The system leader key generates a **base-layer** activation independently of collections, and `buildCollectionBlocks` seeds it into `seenActivators` *before* any collection (KanataConfiguration+BlockBuilders.swift:63–117). So a collection whose momentary activator claims the same base-layer key but routes it to a **different** layer was silently dropped — the user saw the leader's activation work and the collection's vanish, with no explanation ([ADR-039](docs/adr/adr-039-key-conflict-resolution-principles.md) §1).

This threads `leaderKeyPreference` into `RuleCollectionDeduplicator.detectConflicts()` as a synthetic activator claim (slot = `base` + leader key, target = leader's target layer). It then flows through the same `(sourceLayer, input)`-slot logic added in #466.

## Behavior

- **Conflict** — a collection activator on the same base slot as the leader, routing to a *different* layer (e.g. leader `space → nav` + collection `space → window` from base). Surfaced as a "Leader Key" vs collection conflict.
- **Not a conflict** — exact match (default `space → nav` leader vs a nav collection's own `space → nav` activator) is redundant; a same-key chained activator from a *different* source layer (`f → nav` from base vs `f → function` from nav) is a valid chained setup.
- Disabled leader key contributes no claim.

`ConfigurationService` now fetches the leader preference *before* conflict detection so it participates in the gate.

## Regression check

Verified against all 22 seeded collections: only Vim ships a base-layer `space` activator, and it targets `nav` — redundant with the default `space → nav` leader, so it's correctly skipped. No default-enabled collection routes base-`space` to a non-nav layer, so default configs do not begin throwing.

## Tests

Added to `RuleCollectionDeduplicatorTests`: leader-vs-activator conflict surfaced; exact-match redundancy; disabled leader; different key; cross-source-layer chained activator. Full suite green; SwiftFormat 0.61.1 clean on changed files.

Fixes #463

### DIFF
--- a/Sources/KeyPathAppKit/Infrastructure/Config/ConfigurationService.swift
+++ b/Sources/KeyPathAppKit/Infrastructure/Config/ConfigurationService.swift
@@ -241,9 +241,21 @@ public final class ConfigurationService: FileConfigurationProviding {
             )
         }
 
+        // Get leader key preference and trigger mode from PreferencesService on MainActor.
+        // Fetched before conflict detection so the leader key participates in it (#463).
+        let (leaderKeyPref, triggerMode, holdDelayMs) = await MainActor.run {
+            (PreferencesService.shared.leaderKeyPreference,
+             PreferencesService.shared.contextHUDTriggerMode,
+             PreferencesService.shared.contextHUDHoldDelayMs)
+        }
+
         // DETECT CONFLICTS BEFORE DEDUPLICATION
-        // This catches cases where multiple collections map the same key
-        let conflicts = RuleCollectionDeduplicator.detectConflicts(in: collectionsForConfig)
+        // This catches cases where multiple collections map the same key, and where
+        // a collection activator collides with the system leader key (#463).
+        let conflicts = RuleCollectionDeduplicator.detectConflicts(
+            in: collectionsForConfig,
+            leaderKey: leaderKeyPref
+        )
         if !conflicts.isEmpty {
             AppLogger.shared.log(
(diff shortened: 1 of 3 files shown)


malpern:
Good catch — fixed in the latest commit. The leader key consumes its physical key's **base-layer slot** too (not just an activator slot), so I now claim that base slot in the key-overlap pass: a base-layer collection or custom rule mapping the leader key (e.g. default Space leader + a custom Space mapping) surfaces as a `Leader Key` conflict instead of being silently dropped. Verified no seeded collection maps the default Space leader on base, so defaults are unaffected. Added `testDetectsConflictWhenBaseMappingShadowsLeaderKey` + the non-conflict case.

PR #660, opened 2026-05-31
fix: detect duplicate chord group names (#464)

### LINKED ISSUE #464: fix: detect duplicate chord group names
## Problem

Two chord groups with the same name (e.g., both named "navigation") would produce duplicate `defchords` blocks, which kanata rejects. `detectCrossGroupConflicts()` checks for shared keys between groups but not for duplicate group names.

## Current protection

- `kanata --check` CLI validation catches this before config is written
- No pre-flight detection in KeyPath's conflict checking

## Proposed fix

Add a group name uniqueness check to `ChordGroupsConfig` validation, and surface it through the same `MappingConflictInfo` path as other conflicts. Ideally also prevent duplicate names in the chord group editor UI.

## Risk

Low — the UI generates unique names by default, but manual editing or import could create duplicates.

## Related

- #459 — chord group key conflicts

malpern:
## Summary

Part of the ADR-039 conflict-detection cluster.

Each chord group renders to a `(defchords <name> …)` block keyed by its name (KanataConfiguration+Rendering.swift:376/:325). Two groups sharing a name emit duplicate blocks that kanata rejects — previously caught only by `kanata --check`, never surfaced in KeyPath's own pre-flight detection (`detectCrossGroupConflicts()` checks shared *keys*, not names).

This adds `ChordGroupsConfig.detectDuplicateGroupNames()` and surfaces it through the existing chord-group branch of `RuleCollectionDeduplicator.detectConflicts()`, so it flows through the same `MappingConflictInfo` → thrown-error gate as every other conflict, before config generation (ADR-039 §2).

## Behavior

- **Conflict** — two groups in the same `ChordGroupsConfig` with the same exact name.
- Names compared **case-sensitively** (kanata identifiers are case-sensitive; `Nav` and `nav` are distinct).
- Unique names do not conflict. Verified the seeded `benVallackPreset` (groups "Navigation" + "Editing") is unaffected.

## Scope / follow-up

Scoped to within a single `ChordGroupsConfig` (the issue's case). A UI-authored group name colliding with a **preserved hand-written** `(defchords …)` block is a separate, pre-existing gap (different type and rendering source — `ChordGroupConfig` parsed from manual config) and is worth a follow-up; it's not introduced by this change.

## Tests

- `ChordGroupsConfigTests`: detects duplicate names, ignores unique names, case-sensitivity.
- `RuleCollectionDeduplicatorTests`: duplicate names surface via `detectConflicts`; unique names don't.

Full suite green; SwiftFormat 0.61.1 clean on changed files.

Fixes #464

### DIFF
--- a/Sources/KeyPathAppKit/Models/ChordGroupsConfig.swift
+++ b/Sources/KeyPathAppKit/Models/ChordGroupsConfig.swift
@@ -67,11 +67,15 @@ public struct ChordGroupsConfig: Codable, Equatable, Sendable {
     /// two groups with the same name produce duplicate blocks that kanata rejects.
     /// Names are compared exactly (kanata identifiers are case-sensitive).
     ///
+    /// Only groups that actually render are counted: a group with no enabled
+    /// chords is skipped by the renderer, so it cannot collide — counting it would
+    /// flag a false conflict for draft/empty groups.
+    ///
     /// - Returns: One entry per duplicated name, with the number of groups using it,
     ///   sorted by name for stable output.
     public func detectDuplicateGroupNames() -> [(name: String, count: Int)] {
         var counts: [String: Int] = [:]
-        for group in groups {
+        for group in groups where group.chords.contains(where: \.isEnabled) {
             counts[group.name, default: 0] += 1
         }
         return counts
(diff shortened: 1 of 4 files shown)


malpern:
Good catch — fixed in the latest commit. `detectDuplicateGroupNames()` now counts only groups that actually render (those with at least one enabled chord), matching `renderUIChordGroupsBlock`'s `guard !enabledChords.isEmpty` skip. Two same-named draft groups where one has no enabled chords no longer false-flag. Added `testNoDuplicateWhenOneGroupHasNoEnabledChords`.

PR #661, opened 2026-05-31
fix: detect duplicate keys within a single app's keymap (#465)

### LINKED ISSUE #465: fix: detect duplicate keys across app-specific keymaps
## Problem

App-specific keymaps (AppKeymaps.json) generate `kp-{key}` aliases in `keypath-apps.kbd`. If the same key appears twice in the same app's keymap, there's no detection — the alias is silently overwritten.

## Current protection

- `AppConfigGenerator` sorts keys and generates switch expressions, so multiple apps can share a key (the switch dispatches per-app). This is by design for the per-app case.
- However, if the same key appears twice in the same app's keymap, there's no detection.

## Proposed fix

Validate that each input key appears at most once per app keymap during generation. The cross-app case (same key, different apps) is handled correctly by the switch expression.

## Risk

Low — the UI generates keymaps programmatically, making duplicates unlikely.

## Related

- #458 — conflict detection infrastructure

malpern:
## Summary

Part of the ADR-039 conflict-detection cluster.

If the same input key appears twice in **one app's** keymap, `generateAliasBlock` emits two switch cases with the same `((input virtual vk_app))` condition in the `kp-<key>` alias. The second case is unreachable, so the user's second mapping is **silently dropped**. kanata accepts the config (a dead switch case is valid), so this slips past `kanata --check` entirely — it was never surfaced.

Cross-app duplicates (same key, different apps) remain valid — that's the whole point of the per-app switch dispatch.

## Change

- `AppConfigGenerator.detectDuplicateKeys(in:)` — returns per-app duplicated keys, compared **lowercased** and only for **enabled** apps, matching exactly what `generateAliasBlock` renders.
- New `AppConfigError.duplicateKeys` case + `AppKeymapDuplicate` struct.
- `generateAndSave` gates on it and throws before writing (detect-and-explain, ADR-039 §1).

## Why a backstop (and why throwing is safe here)

The UI save path already removes any existing override for the same key before appending (`MapperViewModel+AppKeymapIntegration.swift:51`), and Karabiner import replaces by bundle id — so **no normal flow produces a within-app duplicate**. This gate is a backstop for legacy / hand-edited `AppKeymaps.json`. The throw is consistent with the existing `validationFailed` throw in the same `generateAndSave` function (same blast radius / startup behavior).

## Follow-up (out of scope)

`regenerateFromStore()` regenerates **all** apps from the store, so a corrupt duplicate in one app would block regeneration of others (and leave a stale config at startup, logged not surfaced). This is a pre-existing property of the regenerate-all design (already true for `validationFailed`), not introduced here — worth a separate follow-up to isolate per-app failures / self-heal.

## Tests

`AppConfigGeneratorTests`: in-app duplicate detected; cross-app valid (not detected); unique keys; case-insensitive collision; disabled-app ignored.

Full suite green; SwiftFormat 0.61.1 clean on changed files.

Fixes #465

### DIFF
--- a/Sources/KeyPathAppKit/Services/Configuration/AppConfigGenerator.swift
+++ b/Sources/KeyPathAppKit/Services/Configuration/AppConfigGenerator.swift
@@ -11,6 +11,8 @@ public enum AppConfigError: LocalizedError, Equatable {
     case writeFailed(path: String, underlying: String)
     /// No ConfigurationService available for validation
     case validationUnavailable
+    /// One or more apps map the same input key more than once (#465)
+    case duplicateKeys(duplicates: [AppKeymapDuplicate])
 
     public var errorDescription: String? {
         switch self {
@@ -20,6 +22,10 @@ public enum AppConfigError: LocalizedError, Equatable {
             "Failed to write app config to \(path): \(underlying)"
         case .validationUnavailable:
             "Config validation service unavailable"
+        case let .duplicateKeys(duplicates):
+            "Duplicate keys in app keymap(s): "
+                + duplicates.map { "\($0.app) maps \($0.keys.joined(separator: ", ")) more than once" }
+                .joined(separator: "; ")
         }
     }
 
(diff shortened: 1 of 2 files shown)


malpern:
Good catch — fixed. `detectDuplicateKeys` now groups by `virtualKeyName` across all enabled keymaps (exactly how `generateAliasBlock` keys its switch cases), so two `AppKeymap` entries for the same app each mapping the same key are detected, not just within-entry repeats. Added `testDetectDuplicateKeys_SameKeyAcrossDuplicateAppEntries_Detected`.

Issue #462, opened 2026-05-21
fix: detect alias name collisions from key sanitization


malpern:
## Problem

Alias names are generated by sanitizing input keys: replacing `-` and spaces with `_`. Two different keys can sanitize to the same alias name, producing a duplicate that kanata rejects.

Example: a key named `caps-lock` and one named `caps_lock` both become `beh_base_caps_lock`.

## Affected alias generators

All alias name functions use the same sanitization pattern:
- `behaviorAliasName()` → `beh_{layer}_{sanitized}`
- `forkAliasName()` → `fork_{layer}_{sanitized}`
- `actionAliasName()` → `act_{layer}_{sanitized}`
- `deviceSwitchAliasName()` → `dev_{layer}_{sanitized}`

## Current protection

- `deduplicateAliases()` safety net catches duplicates at config generation time (logs error, keeps last)
- `kanata --check` CLI validation catches duplicates before config is written
- `detectConflicts()` checks input keys but not post-sanitization alias names

## Proposed fix

Compare alias names after sanitization in `detectConflicts()`, or validate alias name uniqueness as a dedicated check before rendering the `defalias` block.

## Risk

Medium — unlikely with standard keyboard keys but possible with custom rules or international layouts.

## Related

- #458 — alias dedup safety net and conflict detection
- #459 — chord group key conflicts

PR #665, opened 2026-05-31
feat: inline resolution dialog for save-time mapping conflicts (#460)

### LINKED ISSUE #460: feat: conflict resolution dialog when collections overlap on the same key
## Context

PR #458 added conflict detection that blocks config writes when two collections configure the same key with different hold behaviors (e.g., Home Row Mods and Home Row Layer Toggles both claiming `;`). The error surfaces through the existing `ValidationFailureDialog` with a plain-English explanation of the conflict.

Currently, the user must manually navigate to the Rules tab and disable one of the conflicting collections. This works but isn't ideal UX.

## Proposed behavior

When a collection conflict is detected during config save, show a **conflict resolution dialog** that:

1. **Explains the conflict** — "Home Row Mods and Home Row Layer Toggles both configure `;` with different hold behaviors. Home Row Mods maps hold to right shift; Layer Toggles maps hold to the fun layer."
2. **Presents choices** — let the user pick which collection's behavior to keep for the conflicting keys, or change the key assignments
3. **Resolves inline** — apply the user's choice without requiring them to navigate to the Rules tab

### Possible resolution options

- **Keep Collection A** — disable the conflicting keys in Collection B
- **Keep Collection B** — disable the conflicting keys in Collection A
- **Change key assignments** — open the relevant collection's config to reassign keys
- **Disable one collection entirely** — turn off one of the two collections

## Technical notes

- The structured conflict data is already available: `MappingConflictInfo` contains the input key, layer, collection names, and per-collection hold descriptions
- The `userExplanation` property generates the English description
- The conflict is thrown as `KeyPathError.configuration(.mappingConflicts(...))` from `ConfigurationService.saveConfiguration()`
- The `ValidationFailureDialog` currently shows the error — this would be a more targeted dialog that replaces or supplements it for mapping conflicts specifically

## Related

- #458 — conflict detection and error surfacing (merged)
- #459 — within-collection chord group key conflicts (separate bug)

malpern:
## Summary

Closes the loop on the ADR-039 detect-and-explain cluster (#462–#466). Those PRs made KeyPath **detect** save-time key conflicts and throw instead of silently dropping a mapping — but the user only saw an error and had to hand-navigate to the Rules tab to disable a collection and retry. This adds an **actionable resolution dialog**.

When a save fails with `mappingConflicts` and every conflicting party is a real, enabled collection (e.g. Home Row Mods vs Home Row Layer Toggles both claiming `;`), the user gets:

```
                    Rule Conflict
  Home Row Mods and Home Row Layer Toggles both configure
  the ";" key with different behaviors. A key can only have
  one action assigned.
      Disable one to resolve. (re-enableable later)

   [ Disable Home Row Mods ]   [ Disable Home Row Layer Toggles ]
```

Pick one → that collection is disabled → **the save automatically retries and succeeds**. No tab-hunting.

## Behavior / scope

- **Actionable only when both parties are real toggleable collections.** Synthetic conflicts — chord-group names (#464), alias key-pairs (#462), the `"Leader Key"` preference (#463) — aren't toggleable, so they fall back to a plain explanation (no misleading button). This is enforced by `resolvableCollectionConflict`, which requires every party to map by exact name to an enabled collection (≥2 distinct).
- **Central interception**, in `RuleCollectionsManager.regenerateConfigFromCollections`'s catch block, so every save path that funnels through regenerate is covered.
- **Disable-and-retry** is depth-guarded and terminates (disabling shrinks the enabled set, so conflicts monotonically decrease).
- The toggle-time rule-conflict dialog still fires first for collection-enable conflicts; this covers conflicts that bypass it (pack installs, restores, structural detections).

## Implementation

Mirrors the proven rule-conflict plumbing 1:1: `RuntimeCoordinator` pending state + `CheckedContinuation` (`promptForMappingConflict`/`resolveMappingConflict`), threaded through `KanataUIState` → `KanataViewModel`, presented as a new `MappingConflictResolutionDialog`.

## Verification

- **Unit-tested**: `SaveTimeConflictResolutionTests` (7 cases) covers the resolvability decision — real-collection vs synthetic/leader/disabled/single/mixed/empty. Full suite green; SwiftFormat 0.61.1 clean on changed files.
- **Needs real-app check** (can't be unit-tested): the sheet appearing, the choice resuming the save, and the retry completing in the running app.

## Follow-up (out of scope)

The inline disable bypasses the pack-ownership check — consistent with the existing toggle-time `.keepNew` path, but worth filtering pack-managed collections from the options later.

Fixes #460

### DIFF
--- a/Sources/KeyPathAppKit/Managers/RuntimeCoordinator.swift
+++ b/Sources/KeyPathAppKit/Managers/RuntimeCoordinator.swift
@@ -111,6 +111,10 @@ public class RuntimeCoordinator: SaveCoordinatorDelegate {
     var pendingRuleConflict: RuleConflictContext?
     var conflictResolutionContinuation: CheckedContinuation<RuleConflictChoice?, Never>?
 
+    // Save-time mapping conflict resolution state (#460)
+    var pendingMappingConflict: MappingConflictContext?
+    var mappingConflictContinuation: CheckedContinuation<UUID?, Never>?
+
     /// Rule collections are now managed by RuleCollectionsCoordinator
     var ruleCollections: [RuleCollection] {
         get { ruleCollectionsCoordinator.ruleCollections }
@@ -434,6 +438,9 @@ public class RuntimeCoordinator: SaveCoordinatorDelegate {
         }
         ruleCollectionsManager.onConflictResolution = { [weak self] context in
             await self?.promptForConflictResolution(context)
+        }
+        ruleCollectionsManager.onMappingConflictResolution = { [weak self] context in
+            await self?.promptForMappingConflict(context)
         }
         // Note: onActionURI callback not needed - RuleCollectionsManager.handleActionURI()
         // already dispatches to ActionDispatcher. Setting this would cause double dispatch.
@@ -1090,7 +1097,8 @@ public class RuntimeCoordinator: SaveCoordinatorDelegate {
             saveStatus: saveStatus,
 
             // Rule conflict resolution
-            pendingRuleConflict: pendingRuleConflict
+            pendingRuleConflict: pendingRuleConflict,
+            pendingMappingConflict: pendingMappingConflict
         )
     }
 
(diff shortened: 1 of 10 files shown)


malpern:
Good catch — fixed. `tryResolveMappingConflict` now snapshots `ruleCollections` before disabling and restores it whenever the retry doesn't return success, so the in-memory disable is atomic (either it leads to a persisted save or it's fully undone). Nested resolutions unwind correctly since each recursion level restores its own snapshot.

PR #667, opened 2026-05-31
fix: detect collection-activator vs mapping key conflicts

malpern:
## Summary

Surfaced by real-app testing of the #460 dialog: enabling **Home Row Mods** + **Home Row Arrows** produced no warning, yet they genuinely conflict on `f`. The live config proved it — both `beh_base_f` (Home Row Mods: hold `f` = Cmd) and `layer_home-arrows_f` (Home Row Arrows' activator) were generated, but the base layer bound `f` to `@layer_home-arrows_f`, **silently dropping Home Row Mods' hold-`f`** (ADR-039 §1 violation).

## Root cause

A collection's **momentary activator** occupies its key's slot on its *source* layer, but the detection passes only cross-checked:
- mapping vs mapping (the `claimedKeys` pass)
- activator vs activator (#466)
- leader vs activator/mapping (#463)

**collection-activator vs another collection's mapping** was covered by none. Home Row Mods has no activator (its mods are dual-role mappings); Home Row Arrows' activator is `f` — so the activator pass (needs two activators) and the mapping pass (only sees mappings, not activators) both miss it.

## Fix

Track which collections map each `(input, layer)` slot (`mappingOwners`, mappings only), and in the activator pass flag when an activator's `(sourceLayer, key)` slot is also mapped by a **different** collection. Lives in the shared `RuleCollectionDeduplicator.detectConflicts`, so it flows through the `mappingConflicts` gate that **both** the GUI dialog (#460) and the CLI apply path use. Disjoint from the activator-vs-activator and leader passes (`mappingOwners` excludes activators and the leader), so no double-reporting.

## Tests

4 new cases in `RuleCollectionDeduplicatorTests` (31 total, green in 7ms): the Home-Row-Mods-shaped conflict + non-fire for different key, different source layer (valid chained), and same-collection.

**Note on local full suite:** the full `swift test` deadlocked locally (a system-probing integration test colliding with the live KeyPath/kanata on this machine — the pure-logic deduplicator tests run in milliseconds). Relying on CI `build-and-test` for the authoritative full-suite run in its clean environment.

### DIFF
--- a/Sources/KeyPathAppKit/Models/RuleCollectionDeduplicator.swift
+++ b/Sources/KeyPathAppKit/Models/RuleCollectionDeduplicator.swift
@@ -20,6 +20,10 @@ enum RuleCollectionDeduplicator {
         }
 
         var claimedKeys: [InputKey: [ClaimInfo]] = [:]
+        // Which collections place a regular mapping on each (input, layer) slot.
+        // Mappings only — NOT activators or the leader — so it can be cross-checked
+        // against activator placements without double-reporting.
+        var mappingOwners: [InputKey: Set<String>] = [:]
 
         for collection in collections where collection.isEnabled {
             let mappings = KanataConfiguration.effectiveMappings(for: collection)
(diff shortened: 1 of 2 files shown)


malpern:
Good catch — fixed. `mappingOwners` is now keyed by collection id → name, and the activator self-filter excludes by `collection.id` (names still shown in the surfaced conflict). Two distinct collections sharing a title no longer cancel each other. Added `testActivatorVsMappingConflictDetectedWhenCollectionsShareName`.

Issue #289, opened 2026-03-15
Expand LogicalKeymap coverage for international layouts


malpern:
## Problem

KeyPath currently has 8 LogicalKeymaps: QWERTY, Colemak, Colemak-DH, Dvorak, Workman, Graphite, AZERTY, and QWERTZ. This means the overlay shows US QWERTY labels for keyboards using Japanese, Korean, Portuguese, Nordic, Russian, and other locale-specific input methods.

The physical layouts exist (macbook-jis, macbook-korean, macbook-abnt2, macbook-iso) but there are no matching LogicalKeymaps to relabel keys for those locales.

## Current State

| Locale | Physical Layout | Logical Keymap | Gap |
|--------|----------------|---------------|-----|
| Japanese | macbook-jis | None | Overlay shows QWERTY labels, not JIS symbols |
| Korean | macbook-korean | None | No Hangul/English toggle label support |
| Brazilian | macbook-abnt2 | None | No ABNT2-specific labels |
| Nordic (NO/SE/DK/FI) | macbook-iso | None | No ø/å/ä labels |
| Russian | None | None | No Cyrillic labels |
| French | macbook-iso | AZERTY (exists) | Covered |
| German | macbook-iso | QWERTZ (exists) | Covered |

## Scope

- Add LogicalKeymaps for Japanese, Korean, Brazilian Portuguese, and Nordic layouts at minimum
- Consider Russian, Spanish, Italian, and other high-demand locales
- Multi-legend rendering (shift labels, hiragana sublegends) is a separate feature — this issue is about primary key labels only
- Input source detection expansion (beyond Japanese IME) is also separate

## Related

- #281 — JIS position table (merged)
- #283 — Keymap fetching improvements (merged)
- QMK locale alias table — resolves import keycodes but doesn't add LogicalKeymaps

## Priority

Medium — the overlay is usable without locale-specific labels (keyCodes and Kanata configs are correct), but the display is confusing for international users.

malpern:
Closing — after implementing this (PR #669, Norwegian + Russian) and reviewing, the static-LogicalKeymap approach is the **wrong mechanism** for international layouts, and the real display need is **already covered**.

[...]

PR #669, opened 2026-05-31
feat: add Norwegian and Russian logical keymaps (part of #289)

malpern:
## Summary

The overlay ships *physical* layouts for international keyboards (JIS, ISO, etc.) but had no matching *logical* keymaps — so users on those keyboards saw **US-QWERTY labels** on the overlay. This adds the first two international keymaps:

- **Norwegian (Nordic):** QWERTY letter arrangement with **Ø** on `;`, **Æ** on `'`, **Å** on `[` (right of P), and `-` in the US `/` slot. Verified against [kbdlayout.info/KBDNO](http://kbdlayout.info/KBDNO/).
- **Russian:** the standard **ЙЦУКЕН (JCUKEN)** Cyrillic layout, with х/ъ/э/ё on the bracket/apostrophe/grave keys. Verified against [JCUKEN (Wikipedia)](https://en.wikipedia.org/wiki/JCUKEN).

Both follow the existing 30-key-block + `extraLabels` model (same as AZERTY/QWERTZ) and are registered in `internationalLayouts`/`all`, so they appear in the overlay keymap picker's International section (`OverlayInspectorPanel+Keymaps.swift:162`).

## Scope

**Part of #289 — does not close it.** Norwegian + Russian are the clean, accurately-encodable wins (pure letter/positional layouts). The remaining listed locales — **Japanese JIS, Korean, Portuguese** — need symbol/kana/jamo label sets (a different kind of data) and are left for a follow-up.

Icons fall back to the placeholder keyboard glyph (no Norwegian/Russian SVG yet) — adding artwork is a follow-up; functionality (correct labels) is unaffected.

## Tests

`InternationalLogicalKeymapTests` (5 cases): registration in `find`/`internationalLayouts`/`all`, plus per-key label assertions for both layouts (core block + the extra Æ/Å/х/ъ/э/ё keys). Build green; SwiftFormat 0.61.1 clean.

### DIFF
--- a/Sources/KeyPathAppKit/Models/LogicalKeymap.swift
+++ b/Sources/KeyPathAppKit/Models/LogicalKeymap.swift
@@ -41,7 +41,9 @@ struct LogicalKeymap: Identifiable {
         graphite,
         // International layouts
         azerty,
-        qwertz
+        qwertz,
+        norwegian,
+        russian
     ]
 
     /// Alternative/ergonomic layouts (excluding QWERTY)
@@ -56,7 +58,9 @@ struct LogicalKeymap: Identifiable {
     /// International layouts
     static let internationalLayouts: [LogicalKeymap] = [
         azerty,
-        qwertz
+        qwertz,
+        norwegian,
+        russian
     ]
 
     /// Look up a static keymap by ID. Returns nil for unknown IDs and for `systemId`.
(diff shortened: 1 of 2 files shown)


malpern:
Closing — the P1 review finding is correct and disqualifying. Selecting a LogicalKeymap doesn't just relabel the overlay; `KeymapMappingGenerator.generateMappings` (KeymapMappingGenerator.swift:41) turns the keymap's `coreLabels` into real kanata `.keystroke` remaps. So selecting **Russian** would generate `q → keystroke("й")` for all 30 keys, and Cyrillic/`ø`/`å`/`æ` aren't valid kanata key tokens — it breaks config generation rather than just changing labels.

Root issue: this mechanism conflates *overlay labeling* with *active key remapping*, which only works for ASCII-Latin alternative layouts (Colemak/Dvorak/AZERTY). International physical-keyboard **display** is already handled by the dynamic **System** keymap (reads the macOS input source via UCKeyTranslate, `generateMappings` returns `[]` for it — no remap). So #289 needs a design decision (decouple label-only keymaps from remapping, or treat it as covered by System), not a data-only PR. Reverting; #289 stays open for that discussion.

PR #781, opened 2026-06-07
Fix HRM insights capability state

### LINKED ISSUE #768: Home Row Mods HRM Insights says Supported while telemetry actions are disabled
## Summary
The Home Row Mods preferences sheet can show HRM Insights as `Supported` while both `Start 60s Calibration` and `Refresh Stats` are disabled. There is no visible or accessible explanation for why the actions are unavailable.

## Evidence from QA pass
Computer Use opened `keypath://settings/rules/home-row-mods` and read the preferences sheet. In the HRM Insights section:

- `home-row-mods-hrm-availability-badge` value: `Supported`
- `home-row-mods-hrm-start-calibration-button`: disabled
- `home-row-mods-hrm-refresh-stats-button`: disabled
- The panel still showed live decision stream entries, so trace data was present.

The code path explains the mismatch: `availability` can become `.supported` from trace events, while the buttons are gated by `supportsHrmStats` (`advertisedCapabilities.contains("hrm-stats")`). A trace-only state therefore looks supported but cannot run stats/calibration actions.

## Expected
The UI should distinguish trace-only support from full stats/calibration support, or explain why the disabled actions are unavailable.

## Suggested acceptance criteria
- If only HRM trace events are supported, the badge or helper text says something like `Trace only` rather than `Supported`.
- Disabled calibration/stat buttons expose an accessibility value or hint explaining the missing capability.
- If stats support should be available whenever trace events are received, fix capability detection so `supportsHrmStats` is set consistently.

malpern:
## Summary
- distinguish HRM trace-only support from full stats/calibration support
- explain disabled HRM stats/calibration actions in the Home Row Mods UI and accessibility hints
- document local Swift build/test performance guidance and Computer Use as preferred UI QA in AGENTS.md and CLAUDE.md

Fixes #768

## Validation
- swift test --filter HrmObservabilityServiceTests
- python3 Scripts/check-accessibility.py
- git diff --check
- swift build via pre-push hook
- quick-deploy + Computer Use verification before commit

Note: a previous full swift test run hit an unrelated timeout/hang in ErrorHandlingTests.testConcurrentConfigurationOperations; targeted coverage for this change is green.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -35,6 +35,26 @@ Notes:
   `poltergeist start`, edit, then `poltergeist wait keypath`. Stop it before
   release work, helper/service work, broad tests, or parallel agents.
 
+### Swift Build/Test Performance
+
+- Prefer targeted tests while iterating:
+  `swift test --filter <TestClassOrMethod>`
+- Run full `swift test` before PR/merge, release-candidate work, or after
+  broad/shared changes.
+- Avoid running broad Swift builds/tests concurrently across worktrees. Before
+  full test, release-candidate, or expensive build work, check for active Swift
+  compiles:
+  `pgrep -fl 'swift-test|swift-build|swift-frontend|swift-driver'`
+- If another worktree/agent is compiling and you must continue, reduce local
+  parallelism:
+  `swift test --jobs 4`
+  `swift build --jobs 4`
+- Do not delete `.build` or run clean builds unless diagnosing cache/build-state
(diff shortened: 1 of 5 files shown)


malpern:
Review note: the CLAUDE.md Linear section removal is intentional. This repo/thread is using GitHub only, so the stale Linear workspace-switching guidance was replaced with the current UI automation guidance.

PR #785, opened 2026-06-07
[codex] Preserve quick-deploy build failure logs

### LINKED ISSUE #784: quick-deploy should preserve full build diagnostics when swift build fails
## Problem

While validating PR #783 for issue #767, `./Scripts/quick-deploy.sh` repeatedly failed during its `swift build` step with only the tail end of a `swift-frontend` stack trace:

```text
🔨 Building...
❌ Build failed
37 swift-frontend ... swift::performFrontend(...)
38 swift-frontend ... swift::mainEntry(...)
39 dyld ... start ...
```

The script writes build output to a temporary log, prints only `tail -3`, then deletes the log. That makes real failures hard to diagnose, especially when the useful compiler diagnostic is earlier in the log.

In this case, direct `swift build` and the equivalent build command succeeded outside `quick-deploy.sh`, while `quick-deploy.sh` still failed before mutating `/Applications/KeyPath.app`. The missing retained log prevented a precise root cause.

## Suggested Fix

- Preserve the full build log on failure, preferably under `.build/logs/quick-deploy/<timestamp>.log` or `.build/quick-deploy-last-build.log`.
- Print the retained log path in the failure message.
- Print a larger and more useful excerpt, such as the first Swift error block plus the final 80-120 lines.
- Consider including build environment details that affect reproducibility: `CLANG_MODULECACHE_PATH`, `SWIFT_MODULECACHE_PATH`, module cache flags, current branch, and build ID.

## Acceptance Criteria

- When `swift build` fails inside `quick-deploy.sh`, the full build log remains available after the script exits.
- The terminal output tells the developer exactly where that log is.
- The failure output includes enough context to distinguish source diagnostics from compiler crashes.
- The script still cleans up deploy locks and restores the launch agent as it does today.

## Context

This came out of #767 / PR #783. It did not block the source fix, but it blocked live installed-app Computer Use smoke validation.

malpern:
## Summary

Fixes #784.

`quick-deploy.sh` now preserves full Swift build diagnostics when a build fails instead of deleting the temporary build log after printing only the final stack frames.

Changes:

- writes quick-deploy build logs under `.build/logs/quick-deploy/build-<build_id>.log`
- adds a build-log header with branch, commit, build ID, module cache paths, and module cache flags
- prints the retained log path on failure
- prints the first relevant diagnostic block plus the final 120 log lines
- keeps successful build logs temporary by deleting them after the build/bin-path steps pass
- adds a lint contract so the failure-diagnostics behavior does not regress

## Validation

- `bash -n Scripts/quick-deploy.sh`
- `TEST_FILTER='DeploymentScriptContractTests' ./Scripts/run-tests-safe.sh` — 3 passed
- `swift build`
- `git diff --check`
- pre-push build hook — passed before push

### DIFF
--- a/Scripts/quick-deploy.sh
+++ b/Scripts/quick-deploy.sh
@@ -38,13 +38,15 @@ MODULE_CACHE_FLAGS=(-Xcc "-fmodules-cache-path=$MODULE_CACHE")
 
 # Build instrumentation
 BUILD_STATS_FILE="$PROJECT_DIR/.build/build-stats.log"
+BUILD_LOG_DIR="$PROJECT_DIR/.build/logs/quick-deploy"
+BUILD_LOG_RETENTION_DAYS="${KEYPATH_QUICK_DEPLOY_LOG_RETENTION_DAYS:-7}"
 LOCK_FILE="$PROJECT_DIR/.build/quick-deploy.lock"
 BUILD_ID="$(date +%s%N | cut -c1-13)"  # Millisecond timestamp as build ID
 
 cd "$PROJECT_DIR"
 
 # Ensure .build directory exists for stats
-mkdir -p "$PROJECT_DIR/.build" "$MODULE_CACHE"
+mkdir -p "$PROJECT_DIR/.build" "$MODULE_CACHE" "$BUILD_LOG_DIR"
 
 # --- Instrumentation Functions ---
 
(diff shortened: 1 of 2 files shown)


malpern:
Addressed the substantive review points in `9e93ab92f`:

- added `KEYPATH_QUICK_DEPLOY_LOG_RETENTION_DAYS` with a default 7-day prune for preserved failed build logs
- replaced the `show-bin-path | tail -1` command substitution with explicit output capture, then `tail -1` only after the Swift command succeeds
- changed the first diagnostic search to start after the `--- swift build output ---` marker so metadata/header lines cannot skew the context block
- expanded the lint contract for retention and `show-bin-path` failure handling

Issue #865, opened 2026-06-10
[Sprint 1.1] Rule dependency awareness: bidirectional prerequisite detection


malpern:
## Status

Tracking-only epic for the KeyPath 1.1 configuration-coherence sequence. Implement through the focused child issues; never ship this epic as one PR.

Core sequence:

1. [#870 — Dependency graph vocabulary and pure builder](https://github.com/malpern/KeyPath/issues/870)
2. [#867 — Per-family provides/requires extraction](https://github.com/malpern/KeyPath/issues/867)
3. [#866 — Forward and reverse detector queries](https://github.com/malpern/KeyPath/issues/866)
4. [#869 — Forward enable/config-edit dialog](https://github.com/malpern/KeyPath/issues/869)
5. [#868 — Reverse disable dialog](https://github.com/malpern/KeyPath/issues/868)

Follow each child’s current labels and hold state. The core model/detector issues may proceed independently of the later human-reviewed dialog work.

## Problem

KeyPath’s existing conflict detector finds collisions where two rules use the same key. It does not detect configuration dependencies where one rule silently relies on behavior supplied by another.

Examples:

- Home Row Layer Toggles references Function, Symbol, Numpad, or Navigation content.
- Quick Launcher in Hyper mode relies on a configured Hyper alias.
- Rules located on or activated from Navigation rely on a functional path into that layer.
- Disabling a provider can leave enabled mappings that silently do nothing.

The Kanata stub-layer safety net keeps many such configurations syntactically valid, but it cannot make missing behavior useful. The product problem is silent failure, not merely parser rejection.

## Outcome

KeyPath analyzes proposed configuration changes in both directions:

- **Forward:** What would be missing after enabling or editing this rule?
- **Reverse:** What would stop working after disabling this provider?

The UI explains concrete consequences and offers a confirmed fix when provider selection is unambiguous.

## Architectural direction

This is a configuration-coherence layer, not a package manager.

### Durable seams

- Model typed capabilities rather than hard-coded rule-to-rule pairs.
- Distinguish:
  - layer content;
  - layer activation/reachability;
  - named aliases such as Hyper.
- Keep capability identity coarse and attach affected keys/configuration as structured evidence.
- Track all known providers separately from currently active providers.
- Normalize open-ended identifiers at graph construction.
- Build deterministically from current configuration; do not persist the graph.
- Keep pure graph types in KeyPathRulesCore.
- Keep family extraction explicit.
- Keep detection, presentation, and mutation separate.
[...]

PR #876, opened 2026-06-11
QA smoke fix: HRL Toggles drops f/j keys to sidestep Home Row Arrows collision

malpern:
Follow-up to [#872](https://github.com/malpern/KeyPath/pull/872). The smoke script as merged was running against the catalog-default HRL Toggles enabledKeys, which assign `f` and `j` to the `nav` layer — but Home Row Arrows is a system-default that also maps `f`. The existing collision detector correctly fires before the stub-deflayer safety net the smoke is meant to exercise can be reached.

## What changed

The Python config-mutation step now drops `f` and `j` from HRL Toggles' enabled keys, and removes them from the layerAssignments + modifierAssignments dicts. The remaining `a`, `s`, `d`, `k`, `l`, `;` keys still point at the `fun` / `sym` / `num` layers, which is what exercises the safety net introduced in [#871](https://github.com/malpern/KeyPath/pull/871).

A real user reaching this combination would either disable Home Row Arrows or reconfigure HRL Toggles. The smoke mirrors the second choice.

## Verification

Ran against the freshly deployed app (`./Scripts/quick-deploy.sh && ./Scripts/qa-hrl-toggles-smoke.sh`):

```
==> case 1: HRL Toggles whileHeld mode, companions disabled
==> case 2: HRL Toggles toggle mode, companions disabled (stub-deflayer safety net)
==> case 3: HRL Toggles toggle mode, companions ENABLED (no stubs needed)
HRL Toggles smoke passed. Restoring original KeyPath config.
```

All three cases pass. Closes the loop on the safety-net being live and verifiable in an installed build.

## Bonus finding worth noting

This is a good signal for the [#865](https://github.com/malpern/KeyPath/issues/865) sprint epic: the existing collision detector is already a first line of defense against the HRL+Arrows class of misconfiguration, and its message ("Either change the key assignments so they don't overlap, or disable one") is exactly the kind of actionable copy the new prerequisite dialogs should mirror.

### DIFF
--- a/Scripts/qa-hrl-toggles-smoke.sh
+++ b/Scripts/qa-hrl-toggles-smoke.sh
@@ -59,6 +59,14 @@ assert_not_contains() {
 # Apply a desired enabled-state to a set of collections and (optionally)
 # flip HRL Toggles' toggleMode. Identifies collections by name to stay
 # robust against UUID drift.
+#
+# Home Row Arrows is a system-default that maps the `f` key to its
+# Home-Arrows layer. The catalog default HRL Toggles config also assigns
+# `f` to its `nav` mapping, which triggers the existing collision detector
+# (good — that's working as intended). To exercise the safety-net code
+# path under test, the script drops `f` from HRL Toggles' enabled keys so
+# the smoke runs past the collision check. Realistic users would either
+# disable Home Row Arrows or reconfigure HRL Toggles' keys.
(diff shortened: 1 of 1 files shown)


malpern:
Duplicate of #875 — both got created by parallel background jobs. Keeping #875 (slightly cleaner formatting).

PR #880, opened 2026-06-11
QA: complete the release-readiness smoke suite (5 scripts + shared lib)

malpern:
Closes ship-gate 4 of the 1.0 release plan: **all 6 family smoke scripts now pass against the installed app.** Follows the template proven by [#872](https://github.com/malpern/KeyPath/pull/872) / [#875](https://github.com/malpern/KeyPath/pull/875).

## What's in this PR

**Shared lib** [`Scripts/lib/qa-smoke-lib.sh`](Scripts/lib/qa-smoke-lib.sh) — extracts the backup/restore/assert/apply plumbing the HRL script established. `smoke_init` backs up the user's real config and registers an EXIT trap that restores it on success *or* failure, so every script is safe on a real machine. `config apply` validates with the bundled kanata, so each successful apply doubles as a syntax check.

**Five new smokes:**

| Script | Covers |
|---|---|
| `qa-auto-shift-smoke.sh` | timeout variants (distinctive 173/247ms values anchor digit assertions), protectFastTyping on/off, reduced key set |
| `qa-launcher-smoke.sh` | holdHyper+hold (asserts `layer-while-held launcher`), holdHyper+tap, leaderSequence |
| `qa-leader-key-smoke.sh` | all 4 singleKeyPicker presets (space/caps/tab/grv), validated against presetOptions |
| `qa-custom-rule-smoke.sh` | CLI rule commands end-to-end: simple remap, tap-hold dual-role, removal |
| `qa-window-snapping-smoke.sh` | enable + standard convention (`window:left/right`, `act_window_*`) |

## Notes from empirical runs

- Leader preset output is `grv`, not `grave` — script validates requested presets against `presetOptions` before applying so a future preset rename fails loudly.
- `rule add`/`rule remove` mutate `CustomRules.json`; the `.kbd` only regenerates on `config apply` — documented in the script.
- Window Snapping's vim convention is deliberately not exercised here: flipping `windowKeyConvention` in JSON alone doesn't regenerate stored mappings (the UI does that), so a JSON-only flip would assert against stale mappings. Vim-convention coverage lives in `PerOptionMatrixTests` at the unit level.

## Verification

All 6 (these 5 + HRL Toggles) pass sequentially against `/Applications/KeyPath.app` with current master deployed:

```
qa-hrl-toggles-smoke:        PASS
qa-auto-shift-smoke:         PASS
qa-launcher-smoke:           PASS
qa-leader-key-smoke:         PASS
qa-custom-rule-smoke:        PASS
qa-window-snapping-smoke:    PASS
```

### DIFF
--- a/Scripts/lib/qa-smoke-lib.sh
+++ b/Scripts/lib/qa-smoke-lib.sh
@@ -23,15 +23,40 @@ require_file() {
   fi
 }
 
+# Restore the user's config and VERIFY the restore actually round-tripped.
+# A restore failure or manifest mismatch escalates to a non-zero exit even if
+# the test assertions passed — a damaged user config must never hide behind a
+# PASS (that is exactly how #881 went unnoticed on its first run).
 smoke_cleanup() {
   local status=$?
   if [[ -n "$BACKUP_PATH" ]]; then
-    "$CLI" config restore --json --quiet "$BACKUP_PATH" --reload >/dev/null || {
-      echo "warning: failed to restore KeyPath config from $BACKUP_PATH" >&2
-    }
+    if ! "$CLI" config restore --json --quiet "$BACKUP_PATH" --reload >/dev/null; then
+      echo "ERROR: failed to restore KeyPath config from $BACKUP_PATH" >&2
+      echo "ERROR: your config may be modified — inspect $CONFIG_DIR against $BACKUP_PATH" >&2
+      status=70
+    elif ! verify_restore_manifest; then
+      status=70
+    fi
   fi
-  [[ -n "$TMP_DIR" ]] && rm -rf "$TMP_DIR"
+  if [[ "$status" -eq 0 && -n "$TMP_DIR" ]]; then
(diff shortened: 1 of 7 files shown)


malpern:
**Holding this PR — do not merge yet.** Running the full suite sequentially tonight surfaced [#881](https://github.com/malpern/KeyPath/issues/881): after repeated backup→restore cycles, the live config dir lost its sidecar files (RuleCollections.json, DeviceSelection.json, installed-packs.json, …) and transient collection state leaked through a restore. The smoke assertions themselves are sound — this is about the safety claim ('safe to run on a real machine'), which doesn't hold until #881 is root-caused.

Plan: diagnose #881 Thursday (fs_usage repro; sync-client interference is the prime suspect), then harden `qa-smoke-lib.sh` to verify the restored item list against a backup manifest and fail loudly on mismatch, then merge this with that guard included. Will also fold in the review feedback (migrate the HRL script to the lib, deepen the Leader Key + Launcher assertions).

PR #885, opened 2026-06-11
Bump kanata engine: fast stuck-key recovery after VHID reconnect (MAL-57 Layer 2)

malpern:
## Summary

Submodule bump pulling [malpern/kanata#26](https://github.com/malpern/kanata/pull/26) into the bundled engine — MAL-57 Layer 2 (Layer 1 was #882). When the kanata↔VHID-daemon socket dies mid-keystroke, a lost key Release no longer autorepeats for ~3 seconds:

1. **Stable per-PID client socket path** — the daemon keys the DriverKit virtual-keyboard *instance* by this path, so a reconnect reuses the existing keyboard instead of spawning a second one while the stuck instance repeats until GC
2. **`virtual_hid_keyboard_reset` on keyboard-ready rising edge** — clears stuck keys as soon as the connection recovers; held keys survive because kanata's reports carry absolute full state
3. **Daemon heartbeat detection 3s → 1s** — matches the daemon's own per-client ping rate

Worst-case burst ~3s → ~1.5s, typical <1s. Also cuts daemon stdout log volume ~95% (ready-status logged on transitions only — it was 183MB of once-per-second spam).

Implemented by vendoring karabiner-driverkit 0.3.1 inside the fork (`[patch.crates-io]`); full causal chain in `docs/bugs/MAL-57-duplicate-keypresses.md`.

## Test plan

- kanata `cargo test` green on the fork branch; `cargo build` + release engine build (`Scripts/build-kanata.sh`, cmd+tcp_server, aarch64) signs and passes binary smoke test
- Post-deploy manual verification: `sudo kill -STOP` the VHID daemon ~4s while holding a key, `kill -CONT`, confirm the repeat burst is bounded (~1s) and held modifiers survive; confirm `virtual_hid_keyboard_reset sent after ready transition` in the kanata log


malpern:
Addressing the review:

**#1 — Fast-forward proof (blocking):** `keypath/bundled` advanced `49b856a3b → 6a05a4a3f` via GitHub's merge of malpern/kanata#26 — not a direct push. Verified locally:
- `git log --oneline 49b856a..6a05a4a` → exactly the merge commit + the one feature commit
- `git rev-parse 6a05a4a^1` → `49b856a3b...` (old tip is the merge's first parent)
- `git merge-base --is-ancestor 49b856a 6a05a4a` → true

No commits lost; the bundled lineage is strictly append-only.

**#3 — Held-key invariant reference (blocking):** the guarantee is stronger than "reset is followed by a full-state report" — *every* report kanata posts is full-state by construction. The `send_key` template at the top of [`driverkit/c_src/driverkit.cpp`](https://github.com/malpern/kanata/blob/6a05a4a3fea117eb1d240d9c77492fb9ed02f7d1/driverkit/c_src/driverkit.cpp#L4-L15) maintains a persistent `keyboard.keys` set (insert on press, erase on release) and posts the **entire report object** on every event. So there is no race to lose: after `virtual_hid_keyboard_reset`, the next key event of any kind re-asserts the complete held-key set atomically in one report.

**#2 — Heartbeat rationale:** 1000ms was chosen to mirror the daemon's own per-client `set_server_check_interval(1000ms)` in `virtual_hid_device_service_clients_manager.hpp` — the daemon already requires clients to be responsive at that cadence, so the client expecting the same of the daemon adds no new tolerance class. The reviewer is right that detection fires 3× sooner under starvation — that is the intent, and it is safe *because of the other two changes in #26*: with the stable per-PID socket path a reconnect reuses the same daemon-side keyboard instance (no ghost-keyboard spawn), and the reset makes each reconnect self-healing. Reconnects are now cheap; staying connected to a dead daemon for 3s is what was expensive. Layer 1 (#882, ProcessType=Interactive) independently reduces the miss rate.

**#4 — Load-test bound:** agreed; the post-deploy verification will include a compile-load run alongside the clean `kill -STOP` test, and results will be recorded in `docs/bugs/MAL-57-duplicate-keypresses.md`.

**Layer 3 tracking:** filed as malpern/kanata#27.

PR #891, opened 2026-06-11
Fix double-fulfill crash in testSingleKeyCaptureLifecycle (unblocks full-coverage lane)

malpern:
The first scheduled `full-coverage` run (M0.1, run 27338726229) crashed the XCTest runner with "API violation — multiple calls made to -[XCTestExpectation fulfill]", thrown while `KindaVimStateAdapterTests` was executing but with the stack pointing back to `testSingleKeyCaptureLifecycle`: its `startCapture` callback and its `asyncAfter` fallback can both fire, and under the slower instrumented build the loser lands **after the test has finished**, fulfilling a completed expectation and killing the entire suite mid-run (which is why the run produced no coverage numbers).

`testContinuousCaptureLifecycle` in the same file already guards this exact race with a lock-protected one-shot fulfill — it was fixed there once and never applied to its sibling. This PR applies the identical pattern. Test-only; no production code changes.

Beyond fixing the coverage lane, this was a latent suite-killer: any sufficiently slow full `swift test` run could hit it.

https://claude.ai/code/session_01Rso6kv5BgJrBCHxNWH4h7P

---
_Generated by [Claude Code](https://claude.ai/code/session_01Rso6kv5BgJrBCHxNWH4h7P)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,11 +58,18 @@ jobs:
     - name: Build and install kanata fork
       run: |
         mkdir -p build/ci-kanata-cache
-        if [[ -x build/ci-kanata-cache/kanata && -x build/ci-kanata-cache/kanata-simulator ]]; then
-          echo "✅ Restored kanata artifacts from cache"
+        # On the self-hosted runner the workspace persists between runs, so
+        # binaries from a previous run can survive even when actions/cache
+        # missed. Gate the "hit" on a stamp matching the checked-out submodule
+        # SHA — otherwise a stale-pin run poisons every later run with an old
+        # engine (stale kanata-simulator broke RemapEndToEndTests; see #891).
+        EXPECTED_ENGINE_SHA=$(git rev-parse HEAD:External/kanata)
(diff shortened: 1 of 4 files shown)


malpern:
Rebuilt this branch as current master + the single fix commit (cherry-pick of 3e7aa2a3f). The previous `build-and-test` failure was the stale base: `RemapEndToEndTests` failures reproduced only against the old base and pass on current master (verified locally). Both `KeyboardCaptureTests` and `RemapEndToEndTests` pass locally on the rebuilt branch. Babysitting CI to merge.

malpern:
Root cause of the lane failure found: the original commit accidentally carried an **External/kanata submodule downgrade** (6a05a4a → 49b856a) from its stale worktree — silently reverting the MAL-57 Layer 2 engine bump (#885) and breaking `RemapEndToEndTests` in the full lane (CI builds the old simulator from the reverted pin; locally the newer submodule checkout masked it). Amended to keep the pin at 6a05a4a. Also folded in the Neovim pack 19-binding fix for `testPackBindingsMatchCollectionMappings` (full-lane-only invariant that #893's standard run never executed). Polling CI.

PR #901, opened 2026-06-11
Redirect diagnostic/support writes to a temp sandbox during tests

malpern:
## Problem

Four services wrote to real user directories while unit tests ran, risking pollution or eviction of genuine diagnostic data:

- `MainAppStateController.logCrashEvent()` and `KanataDaemonService.logServiceFailure()` appended to `~/Library/Logs/KeyPath/crashes.log`
- `KindaVimTelemetryStore.defaultFileURL` pointed at `~/Library/Application Support/KeyPath/kindavim-telemetry.json`
- `StuckKeyRecoveryService` wrote incident snapshots to `~/Library/Logs/KeyPath/stuck-key-incidents/`, which sits at its 20-file prune cap — a test write silently evicts a real incident file (relevant to MAL-57 evidence)

## Fix

New `AppPaths` utility in KeyPathCore resolves user-scoped directories through one helper that redirects into a per-process temp sandbox (`$TMPDIR/keypath-tests-<pid>/…`) when `TestEnvironment.isRunningTests`. All four services route through it instead of hand-rolling the check. **Production paths are byte-identical to before.** Test detection is resolved once per process (Bundle.allBundles scan is not repeated per access).

`InstalledPackTracker` keeps its existing per-instance UUID redirect — it provides stronger isolation between test instances than the shared sandbox would.

## Tests

- New `AppPathsTestSandboxTests` (7 tests) asserts the redirect for each path, including a sanity test that test detection itself works.
- Verified end-to-end: after running the touched suites, real `crashes.log` mtime and all 20 real incident files are unchanged, while the stuck-key test's snapshot landed in the sandbox.

## Follow-up (out of scope)

`ActivityLogStorage` (KeyPathInsights) has the same bug class; flagged separately.

### DIFF
--- a/Sources/KeyPathAppKit/Services/Kanata/KanataDaemonService.swift
+++ b/Sources/KeyPathAppKit/Services/Kanata/KanataDaemonService.swift
@@ -291,10 +291,10 @@ final class KanataDaemonService {
     }
 
     /// Log service state failures to persistent crash log for later analysis
+    /// (redirected to a temp sandbox during tests via AppPaths).
     private func logServiceFailure(from oldState: ServiceState, reason: String) {
-        let crashLogDir = Foundation.FileManager().homeDirectoryForCurrentUser
-            .appendingPathComponent("Library/Logs/KeyPath")
-        let crashLogPath = crashLogDir.appendingPathComponent("crashes.log")
+        let crashLogDir = AppPaths.logsDirectory
+        let crashLogPath = AppPaths.crashLogFile
 
         // Ensure directory exists
         do {
--- a/Sources/KeyPathAppKit/Services/KindaVim/KindaVimTelemetryStore.swift
+++ b/Sources/KeyPathAppKit/Services/KindaVim/KindaVimTelemetryStore.swift
@@ -128,12 +128,11 @@ final class KindaVimTelemetryStore {
     /// `UserDefaults` key.
     static let optInKey = "kindaVim.telemetryEnabled"
 
+    /// ~/Library/Application Support/KeyPath/kindavim-telemetry.json
+    /// (redirected to a temp sandbox during tests via AppPaths).
     static var defaultFileURL: URL {
-        let support = FileManager.default
-            .homeDirectoryForCurrentUser
-            .appendingPathComponent("Library/Application Support", isDirectory: true)
-            .appendingPathComponent("KeyPath", isDirectory: true)
-        return support.appendingPathComponent("kindavim-telemetry.json")
+        AppPaths.applicationSupportDirectory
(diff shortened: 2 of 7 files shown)


malpern:
Addressed in 86cc94342:

1. **Legacy path grep** — done. The only reference to the old path is `StuckKeyRecoveryServiceTests.testDiagnosticSnapshotsRedirectToTempDirectoryDuringTests`, which uses it purely as a negative assertion (`XCTAssertNotEqual` vs the sandbox), so it remains valid. Nothing relies on a legacy test-mode location (there was none on master — the redirect is new in this PR).
2. **Sandbox cleanup** — added `class func tearDown()` in `AppPathsTestSandboxTests` removing the per-process sandbox. Safe with later suites: every production writer (`logCrashEvent`, `logServiceFailure`, `writeSnapshotToDisk`, `KindaVimTelemetryStore.persist`) creates its directory before writing.
3. **`testSandboxDirectory` as `static let`** — done.
4. **`@MainActor` on the test class** — kept: `KindaVimTelemetryStore` is a `@MainActor` class, so its static `defaultFileURL` is main-actor-isolated and the test must read it from the main actor.
5. **`public` vs `internal`** — kept public: `AppPaths` is used cross-module from KeyPathAppKit, and a public `testSandboxDirectory` lets tests use a plain import; it's documented as test scaffolding.

Nit: the four doc comments already share the same trailing-period form.

Follow-up for `ActivityLogStorage` is already tracked as a spawned task.

malpern:
Both blocking items from the latest review are resolved in 6bdbfe8c2:

1. **Legacy path** — `grep -rn "KeyPathTests/stuck-key-incidents" Tests/ Sources/ Scripts/` returns nothing. The #886 redirect this PR supersedes left no test referencing the old location; the only path assertion is `testDiagnosticSnapshotsRedirectToTempDirectoryDuringTests`, which compares against the *real* directory (negative) and `temporaryDirectory` (positive prefix) — both hold under the new per-PID sandbox, non-vacuously.
2. **InstalledPackTracker callsite comment** — added, explaining the per-instance UUID directory is intentional (parallel tracker instances must not share records, which the shared per-process sandbox would cause).

On the recurring `@MainActor` suggestion: it is required — `KindaVimTelemetryStore` is a `@MainActor` class, so its static `defaultFileURL` is main-actor-isolated and the test must read it from the main actor.

Write-level coverage note: `StuckKeyRecoveryServiceTests` already exercises a real snapshot write through the service (restart path), which lands in the sandbox; combined with the path assertions this covers the regression surface. A crashes.log write-through smoke test is reasonable follow-up polish.

PR #904, opened 2026-06-11
Wizard: proactively surface stale pre-MAL-57 VHID daemon plist with one-click repair

malpern:
## Summary

`ServiceHealthChecker.isVHIDDaemonConfiguredCorrectly()` (ProgramArguments path + `ProcessType=Interactive`, the MAL-57 starvation fix) was only consumed by the repair postflight, where it trivially passes because repair just rewrote the plist. Old installs with a pre-fix plist never migrated unless repair ran for some other reason, leaving them vulnerable to the MAL-57 stuck-key autorepeat under CPU load.

This wires plist-content validation into the wizard's health model:

- **`ServiceHealthChecker.isVHIDDaemonPlistPresentButMisconfigured()`** — single-read, presence-gated content check (shared with `isVHIDDaemonConfiguredCorrectly` via one content validator using `PlistGenerator.vhidDaemonPath`). A missing plist stays attributed to service health ("not installed"); an existing-but-stale or unparseable plist counts as misconfigured.
- **`ComponentStatus.vhidDaemonPlistMisconfigured`** — new fact, populated by `SystemValidator` inside the 15s-TTL component-facts cache (no per-cycle main-thread plist I/O; `invalidateCaches()` refreshes it after repairs). Deliberately excluded from `hasAllRequired` (documented): a stale plist still runs day-to-day.
- **`SystemInspector`** emits the previously dormant `.component(.vhidDaemonMisconfigured)` issue — severity **`.warning`** (nudges migration without flipping the app to a failed state; wizard routing and the fix button are identifier-based and severity-agnostic), gated on otherwise-healthy services so it doesn't double-alarm with "VHID Services Unhealthy". Auto-fix: `.installRequiredRuntimeServices`. Existing consumers (`KarabinerComponentsStatusEvaluator`, `WizardRouter`, components-page repair) light up unchanged.
- **`ActionDeterminer`** repair/install plans include the rewrite via a new shared `ComponentStatus.vhidRuntimeServicesNeedRepair`, so the trigger can't drift between repair and install.
- **Stale-helper guard** (review finding, confirmed): a resident pre-fix helper rewrites the old plist and reports success, which would loop the fix forever. `PrivilegedOperationsRouter.installRequiredRuntimeServices()` now verifies the plist after the helper path and falls back to the in-app install (current `PlistGenerator`) so the repair converges.

Review gate findings addressed; notably severity was downgraded from `.error` after verifying `MainAppStateController` counts `.installation`+`.error` issues as blocking (would have shown an app-wide red/failed state for working installs).

## Test plan

- New `ServiceHealthCheckerTests`: misconfigured check for missing / correct / no-ProcessType / wrong-path / corrupt plists (hermetic via `KEYPATH_LAUNCH_DAEMONS_DIR`); fixtures now default to `PlistGenerator.vhidDaemonPath`; singleton health cache cleared in `setUp` (fixes a cross-suite flake).
- New `WizardPureLogicTests`: inspector emits the issue (`.installation`, `.warning`, auto-fix, state stays `.active`); suppressed while services are unhealthy (single alarm, same fix); `ActionDeterminer` includes the rewrite when only the plist is stale.
- Full safe gate: 4,378 tests pass.

**Known coverage gap:** the stale-helper guard in `PrivilegedOperationsRouter.installRequiredRuntimeServices()` (helper reports success → plist still stale → in-app sudo rewrite) has no unit test — it requires mocking a succeeding `HelperManager` alongside the injected `ServiceHealthChecker`. The pure-logic detection, issue gating, and action triggers around it are covered; cache freshness after the fallback is handled by `MainAppStateController.invalidateValidationCooldown()` on wizard close (15s TTL worst case mid-wizard).

### DIFF
--- a/Sources/KeyPathAppKit/Core/PrivilegedOperationsRouter.swift
+++ b/Sources/KeyPathAppKit/Core/PrivilegedOperationsRouter.swift
@@ -95,6 +95,22 @@ public final class PrivilegedOperationsRouter {
         case .privilegedHelper:
             do {
                 try await helperManager.installRequiredRuntimeServices()
+                // Stale-helper guard (MAL-57): a resident helper from before
+                // the ProcessType=Interactive template fix reports success but
+                // rewrites the old plist — the helper binary only refreshes on
+                // process restart, so it can lag the app across an update.
+                // Verify the result and redo the rewrite via the in-app sudo
+                // path (ServiceBootstrapper + current PlistGenerator, with its
+                // own configured-postflight). Must NOT route back through the
+                // helper — sudoInstallRequiredRuntimeServices would, via
+                // InstallerEngine → PrivilegeBroker → this router's
+                // helper-first repairVHIDDaemonServices.
+                if serviceHealthChecker.isVHIDDaemonPlistPresentButMisconfigured() {
+                    AppLogger.shared.warn(
+                        "⚠️ [PrivilegedOperationsRouter] VHID daemon plist still misconfigured after helper install — stale helper template, rewriting via in-app sudo path"
+                    )
+                    try await sudoRepairVHIDServices()
+                }
             } catch {
                 try await sudoInstallRequiredRuntimeServices()
             }
--- a/Sources/KeyPathAppKit/Services/System/SystemValidator.swift
+++ b/Sources/KeyPathAppKit/Services/System/SystemValidator.swift
@@ -47,6 +47,7 @@ public class SystemValidator {
         let vhidInstalled: Bool
         let vhidVersionMismatch: Bool
         let karabinerDriverExtensionEnabled: Bool
+        let vhidDaemonPlistMisconfigured: Bool
         let timestamp: Date
 
         func isFresh(ttl: TimeInterval) -> Bool {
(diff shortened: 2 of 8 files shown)


malpern:
Review addressed (picked up from a parallel session since this PR had stalled at BEHIND):

- **Install-path test added**: `test_determineInstallActions_vhidDaemonPlistMisconfigured_includesInstallRuntimeServices` mirrors the repair-path test, guarding the exact drift `vhidRuntimeServicesNeedRepair` exists to prevent.
- **`FileManager.default`** used in the new `isVHIDDaemonPlistPresentButMisconfigured()`. One note for the record: the rest of ServiceHealthChecker consistently uses `Foundation.FileManager()` (lines 197, 253, 327, 873), so the 'inconsistent with the codebase' premise was inverted — but `.default` is the better API and the right direction for new code.
- **do-catch nit**: left as-is; that formatting came from the pinned swiftformat config, so hand-reverting would churn on the next format pass.

Rebased onto master (now includes #897/#898/#902). `WizardPureLogicTests` + `ServiceHealthCheckerTests` green locally (117 tests).

malpern:
Review findings addressed in the updated commit:

- **Codex P1 (stale-helper retry loops back through the helper)** — confirmed and fixed. The guard now calls `sudoRepairVHIDServices()` (ServiceBootstrapper + current PlistGenerator, with its own configured-postflight) instead of `sudoInstallRequiredRuntimeServices()`, which would have routed back through InstallerEngine → PrivilegeBroker → this router's helper-first `repairVHIDDaemonServices`. The comment in the code documents the trap.
- **Missing `determineInstallActions` test** — added (`test_determineInstallActions_vhidDaemonPlistMisconfigured_includesInstallRuntimeServices`).
- **Single-line do-catch compression** — reverted; the router keeps master's two-line form (that was lint-autofix churn, not intentional).
- **`Foundation.FileManager()` vs `FileManager.default`** — kept as is: `ServiceHealthChecker.swift` consistently uses `Foundation.FileManager()` (see the test-mode plist-existence checks at ~197/253/327), so the file-local convention is the instance form.

PR #936, opened 2026-07-02
Fix ConfigurationServiceTests isolation: inject rule stores into ConfigurationService

malpern:
## Summary

`testCreateInitialConfigWritesDefaultFile` and `testReloadCreatesDefaultConfigWhenMissing` failed on any developer machine whose real rule stores contain conflicting collections (e.g. "Home Row Arrows" + "Home Row Mods" both enabled → `KeyPathError.ConfigurationError.mappingConflicts`). Root cause: `ConfigurationService.createInitialConfigIfNeeded()` read `RuleCollectionStore.shared` / `CustomRulesStore.shared` — the real user config-directory stores — even under XCTest, because the stores are process-global singletons (the tests already used a temp *config* directory, but not temp *stores*).

- `ConfigurationService` now takes the two stores via an internal designated initializer and uses them in `createInitialConfigIfNeeded()`.
- The `public init(configDirectory:)` becomes a convenience init defaulting to `.shared` — signature and behavior unchanged for every production call site, and identical behavior on clean CI (empty temp stores load catalog defaults, exactly what the shared stores return on a clean machine).
- `ConfigurationServiceTests` passes temp-backed stores (existing `testStore(at:)` helpers), making the suite hermetic on any machine.
- New regression test `testCreateInitialConfigReadsInjectedStores` seeds an injected store with a marker collection and asserts initial-config rehydration reads it (not the shared singletons).

## Test plan

- Reproduced the failure pre-fix on a machine with conflicting user collections: `testCreateInitialConfigWritesDefaultFile` fails with `mappingConflicts(inputKey: "f", conflictingCollections: ["Home Row Arrows", "Home Row Mods"])`.
- Post-fix: `swift test --filter ConfigurationServiceTests` → 72/72 pass on that same machine.
- Full safe suite run locally: 1761 passed; the remaining failures (snapshot rendering, CLIPackCRUD/HelperMaintenance full-lane interference) reproduce identically on master on this macOS 27 beta machine and are unrelated.

### DIFF
--- a/Sources/KeyPathAppKit/Infrastructure/Config/ConfigurationService.swift
+++ b/Sources/KeyPathAppKit/Infrastructure/Config/ConfigurationService.swift
@@ -29,6 +29,9 @@ public final class ConfigurationService: FileConfigurationProviding {
     private var fileWatcher: FileWatcher?
     private var observers: [UUID: @Sendable (Config) async -> Void] = [:]
 
+    private let ruleCollectionStore: RuleCollectionStore
+    private let customRulesStore: CustomRulesStore
+
     /// Perform blocking file I/O off the main actor
     private let ioQueue = DispatchQueue(label: "com.keypath.configservice.io", qos: .utility)
     /// Protect shared state when accessed from multiple threads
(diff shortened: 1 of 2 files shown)


malpern:
Re the flagged gap: audited the other suites that can reach createInitialConfigIfNeeded via reload(). Only ConfigurationServiceSavePipelineTests.testReload_ThrowsFileNotFoundWhenMissing constructs with the public init and reloads a missing config, and it deliberately tolerates every outcome (including store-conflict errors), so it can't fail from developer-machine store state. The remaining suites write a config file before reloading or pass explicit collections, so they never hit the store-read path. Keeping this PR scoped to the failing suite; a broader store-injection sweep can ride along if that ever flakes.

Issue #934, opened 2026-07-02
Wizard polish batch: restart messaging, prompt wording, progress indicator, completion moment, error recovery, misc


malpern:
This tracking batch was re-audited against current `master` on 2026-07-14. It should not remain open as a broad implementation target.

## Completed or no longer actionable

- [x] Progress indicator, wizard-open flicker, cosmetic error severity, and validation-loop noise — PR #947.
- [x] Input Monitoring granted-state copy now says the main application can capture keyboard input.
- [x] The obsolete user-facing legacy-helper cleanup prompt is no longer present; legacy cleanup wording remains internal to the privileged command description and logs.
- [x] Login Items approval is explicitly narrated with contextual copy, a screenshot, a concrete toggle instruction, and polling.
- [x] Setup is discoverable from the main-window status control and the command palette; a second Settings-only entry is not required without usability evidence.
- [x] Broad error-recovery request retired: current pages expose concrete recovery actions. Any observed missing recovery path should be filed as its own reproducible defect.

## Split or transferred

- Clean-account grant recognition and restart-required behavior → #747.
- First-success celebration, starter mapping, and panel tour → #954.
- Permission-action vocabulary (`Turn On`, `Add in Settings`, `Open Settings`) → #1181.

## Closeout

The remaining work now has focused owners and priorities. Do not reopen this batch; file a narrow issue with runtime or usability evidence.

malpern:
Codex note from #931 local UI pass (2026-07-04): add this to the wizard polish table/batch.

Input Monitoring page can show `KeyPath.app` with a green check while the subtitle still says `Main application needs permission`. The permission state is correct; the row copy should switch to granted-state wording, e.g. `Main application can capture keyboard input` or similar.

This is polish only; not a #931 blocker.

malpern:
Backlog hygiene checkpoint: PR #947 completed the progress indicator, wizard-open flicker, cosmetic log severity, and repeated validation-log noise items. The remaining unchecked items still require product or UX work. The first-success implementation attempted in PR #966 was closed without merging, so the completion-moment item remains open.

malpern:
Backlog revision (2026-07-13): marked tracking-only. Completed items remain historical; each remaining observed UX problem should be taken as a narrow slice or split into its own issue rather than implemented as one polish batch.

malpern:
[...]

Issue #933, opened 2026-07-02
Wizard: drag-to-authorize parity — KeyPath.app rows (AX + IM) and Full Disk Access lack the drag overlay


malpern:
The OpenAI-style drag-to-add overlay (`DragToAuthorizeController`) is only wired to the **kanata-launcher** rows:

- Input Monitoring: kanata-launcher row → drag overlay (WizardInputMonitoringPage.swift:202); **KeyPath.app row → bare "Turn On" deep link** (line 168)
- Accessibility: same split (WizardAccessibilityPage.swift:220 vs :182)
- Full Disk Access: **no drag UI at all**, just an "Enable" deep-link button (WizardFullDiskAccessPage.swift:171) — and unlike AX/IM it doesn't poll for the grant (only re-checks on didBecomeActive), so it never auto-advances

## Proposal

- Extend the drag overlay to KeyPath.app rows on AX + IM and to the FDA page
- Give FDA the same 1s polling + checkmark-burst auto-advance the other permission pages have
- Give IM a manual "Open Settings" fallback link (it currently only has the drag button; AX has both)

From the 2026-07-02 fresh-install QA review. Related: #749, #931 (IM detection — drag guidance is moot if the app isn't in the Settings list yet).

Issue #932, opened 2026-07-02
Wizard: add 'Welcome to KeyPath' intro page (what it is, value, steps ahead)


malpern:
The wizard currently opens directly on the summary/status page — on a fresh install that's a diagnostic dump of ~8 red issues as the user's first impression (WizardTypes.swift `orderedPages`, wizard auto-launches from App.swift on fresh install).

## Proposal

A welcome page shown once on fresh install, before the status summary:
- What KeyPath is + the value it offers (1-2 sentences, product voice)
- What setup will involve: 3-4 illustrated steps (helper approval, permissions, driver, start service) so macOS's prompts don't feel alarming
- Sets expectations for admin password + System Settings hops
- "Get Started" CTA → existing flow

Goal: Apple-award-level first-run feel. Pairs with the completion/onboarding moment (see polish batch issue).

From the 2026-07-02 fresh-install QA review. Related: #749.

Issue #931, opened 2026-07-02
Permissions: Input Monitoring grant never produced a TCC row the Oracle recognizes — fresh install dead-ends on IM page


malpern:
## Symptom (fresh-install session, 2026-07-02, v1.0.0 RC 812245f0, macOS 27.0 beta)

From 05:17:23 to end of session (6+ minutes), the wizard sat on the Input Monitoring page:

- Oracle reported `IM: unknown` in all 113 consecutive snapshots
- TCC query for `kTCCServiceListenEvent` returned **empty for BOTH com.keypath.KeyPath and kanata-launcher** every time — the row never appeared
- Accessibility, by contrast, worked (after app restart — separate issue)
- This is where the install effectively died; the user gave up and sought help

Note: also reproduce/verify on macOS 26.x — this session was on a macOS 27 beta (see #919), but the detection gap (unknown forever, no user guidance) is a bug regardless.

## Investigate / fix direction

- Determine whether the user's grant flow ever completed (does the OS prompt appear? does the Settings toggle exist for a not-yet-prompting app?) — on modern macOS an app may need to call `IOHIDRequestAccess` first to appear in the IM list at all.
- If the app/kanata-launcher isn't in the list yet, the wizard's guidance ("Turn On" / drag-to-add) points at a list entry that doesn't exist.
- Oracle: distinguish "no TCC row = never prompted" from "denied", and drive the wizard copy accordingly (e.g., trigger the prompt first).
- Add the same restart-detection awareness AX needed.

## Acceptance
A fresh install can complete the IM step: the wizard triggers/detects the grant reliably or gives accurate manual instructions, and never sits at `unknown` indefinitely without guidance.

Part of the 1.0 install-QA tier-0 batch. Related: #919, #747, ADR-001/ADR-006.

malpern:
## Investigation results (2026-07-02 session, macOS 27.0 beta)

[...]

PR #939, opened 2026-07-03
Fix Input Monitoring wizard dead-end: escalate to manual add + keep poll alive (#931)

malpern:
## Summary

Closes the fresh-install **Input Monitoring dead-end** on macOS 26/27 — the spot where the 2026-07-02 QA session "effectively died" (#931). #937 already fixed KeyPath.app's own IM *signal* (`IOHIDCheckAccess`); this fixes the wizard **UX** that was still stranded, plus a root-cause polling-lifecycle bug that would have defeated the fix.

### The problem
Clicking **Turn On** fires `IOHIDRequestAccess`, but on macOS 26/27 tccd can produce **no prompt and no TCC row** — the app never appears in the Input Monitoring list. The page kept showing only a bare "Turn On" button that deep-links to a Settings row which doesn't exist, giving the user no accurate way forward.

### The fix
- **Pure, unit-tested resolver** (`InputMonitoringGuidance`, in `KeyPathWizardCore`) that escalates KeyPath.app's own IM guidance: offer the automatic prompt first, then after a short wait window without a grant switch to explicit manual **"+"-to-add** steps. A `nil` or negative (clock-rewind) elapsed escalates rather than stranding the user in an endless "waiting" state.
- **`InputMonitoringManualFallbackCard`** rendered from that state, so the page never sits unresolved without accurate manual instructions — satisfying the #931 acceptance criterion ("never sits at `unknown` indefinitely without guidance").

### Root-cause polling fix (caught in review)
The escalation relies on periodic re-renders driven by the 1s `permissionSnapshot` poll. But the fast post-prompt poll (`startPermissionPolling`) **cancelled** the durable 1s poll, **stopped writing** `permissionSnapshot`, and **never restarted** it — so after ~5s the page went static: no re-renders (the card could never appear) and no detection of a later *manual* grant (the wizard could never advance). Fixed by extracting a durable `startBackgroundPermissionPolling()`, having the fast poll also refresh the snapshot, and resuming the durable poll when the fast poll exhausts.

## Testing
- New `InputMonitoringGuidanceTests` cover every resolver transition, the wait-window boundary, and the `nil`/negative-elapsed escalation guards.
- ⚠️ Swift build, full `swift test`, `swiftformat`, and `swiftlint` **were not run** here — this change was authored in a Linux container with no Swift/macOS toolchain. Please run the pre-PR gate on macOS.
- Manual first-run QA (#747) still required on a Mac without pre-granted permissions, ideally on a macOS 26 and a macOS 27 machine.

## Scope / follow-ups
Deliberately scoped to the KeyPath.app IM escalation. Still open on #931 and tracked separately:
- kanata-launcher self-registration / daemon-before-IM-step (#931)
- Welcome intro page (#932), drag-to-authorize parity + FDA polling (#933), wizard polish batch (#934)

## Review gate
CLAUDE.md's mandated `/thermo-nuclear-swift-review` command isn't present in this checkout, so the built-in `/code-review` (high effort) was run against the diff instead; all confirmed findings — the fatal dead-poll bug, the clock-rewind edge, and the design-token/comment drift — were addressed.

https://claude.ai/code/session_015fnREsXoC9obKsFkF7eKPB

---
_Generated by [Claude Code](https://claude.ai/code/session_015fnREsXoC9obKsFkF7eKPB)_

### DIFF
--- a/Sources/KeyPathInstallationWizard/UI/Pages/WizardInputMonitoringPage.swift
+++ b/Sources/KeyPathInstallationWizard/UI/Pages/WizardInputMonitoringPage.swift
@@ -463,15 +463,18 @@ public struct WizardInputMonitoringPage: View {
             "🔧 [WizardInputMonitoringPage] Fix button clicked - permission flow starting"
         )
 
-        // Record the attempt so guidance escalates to manual steps if the
-        // automatic prompt never registers a grant (macOS 26/27 dead-end, #931).
-        keyPathRequestAttemptedAt = Date()
-
         // Use automatic prompt via IOHIDRequestAccess()
         guard let permissionRequestService = WizardDependencies.permissionRequestService else {
             AppLogger.shared.log("⚠️ [WizardInputMonitoringPage] permissionRequestService not configured")
             return
         }
+
+        // Record the attempt so guidance escalates to manual steps if the
+        // automatic prompt never registers a grant (macOS 26/27 dead-end, #931).
+        // Set only after the guard so the escalation clock never starts on a path
+        // where no request was actually made.
+        keyPathRequestAttemptedAt = Date()
+
         Task { @MainActor in
             let alreadyGranted = await permissionRequestService.requestInputMonitoringPermission(
                 ignoreCooldown: true
--- a/Sources/KeyPathInstallationWizard/UI/Pages/WizardInputMonitoringPage.swift
+++ b/Sources/KeyPathInstallationWizard/UI/Pages/WizardInputMonitoringPage.swift
@@ -454,6 +454,12 @@ public struct WizardInputMonitoringPage: View {
             // the page keeps refreshing (drives the #931 manual-fallback card and
             // still detects a later manual grant). Without this the page would go
             // static and the escalation could never appear.
+            //
+            // Note: startBackgroundPermissionPolling() begins by cancelling
+            // `permissionPollingTask`, which still points at THIS fast-poll task —
+            // an intentional self-cancel. It is harmless: cancellation only flips
+            // `Task.isCancelled`, and there is no further code or await point in
+            // this closure after the call. Do not "simplify" it away.
             startBackgroundPermissionPolling()
         }
     }
--- a/Sources/KeyPathInstallationWizard/UI/Pages/WizardInputMonitoringPage.swift
+++ b/Sources/KeyPathInstallationWizard/UI/Pages/WizardInputMonitoringPage.swift
@@ -16,6 +16,10 @@ public struct WizardInputMonitoringPage: View {
     @State private var permissionPollingTask: Task<Void, Never>?
     @State private var showSuccessBurst = false
     @State private var permissionSnapshot: PermissionOracle.Snapshot?
+    /// When the automatic `IOHIDRequestAccess` prompt for KeyPath.app was last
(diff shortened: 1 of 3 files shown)


malpern:
Thanks for the review. Addressed the two actionable observations in 31b7c16:

- **Obs #2** — moved the `keyPathRequestAttemptedAt` assignment below the `permissionRequestService` nil-guard, so the escalation clock never starts on a path where no request was actually made.
- **Obs #1** — widened the default wait window from 6s → 12s so a *working* system prompt (which a first-run user may take several seconds to answer) isn't contradicted by "Didn't see a prompt?" while it's still on screen, while still not stranding a genuinely stuck user. Added a test locking in the 12s default.

Obs #3 (two Settings-pointing fallbacks) and #4 (kanata scope) I've left as-is — #3 is a QA-confirm item for #747 and #4 is the intended deferral to the #931 follow-ups. Build/format/lint still need to pass CI since this was authored without a Swift toolchain.

---

malpern:
## On-hardware QA (macOS 27 beta, real Developer-ID build) ✅

Built, signed, and deployed this branch to `/Applications/KeyPath.app` and drove the wizard with Computer Use. Findings:

**Pre-PR gate**
- `InputMonitoringGuidanceTests` (8) pass; installer lane (457 tests) pass.
- swiftformat 0.61.1 clean, no churn. (Local swiftlint 0.63.3 wants unrelated repo-wide `trailing_comma` edits — a version-mismatch, not from this PR; not applied.)

**#931 escalation — reproduced and verified end-to-end.** Because a Developer-ID re-sign preserves the TCC grant *and* the macOS 26/27 kernel HID grant is sticky across `tccutil reset`, the not-granted state was reached via a temporary env-gated debug hook (reverted, never committed). With KeyPath IM forced denied:
- Wizard IM page shows `KeyPath.app ❌ Main application needs permission` + `Turn On`.
- Clicking **Turn On** → System Settings auto-opens (~1.5s) to Input Monitoring, showing a list that does **not** contain KeyPath (exactly the dead-end: no prompt, no TCC row).
- At the ~12s window the **"Didn't see a permission prompt?" manual-add card** appears below the two status rows. The page never sat unresolved.

**Root-cause polling fix — confirmed from the log cadence** after Turn On: ~250ms checks for the ~5s fast-poll window, then checks **resume at ~1s and continue indefinitely** (the durable poll restarting). Without the fix, checks would have stopped at ~5s and the card could never render. This is the mechanism that drives the escalation.

**#937 IM signal** — Oracle reports KeyPath IM via `im-api` (IOHIDCheckAccess), authoritative over TCC; it correctly ignored a `tccutil reset` (still reported granted), which is precisely the macOS 26/27 divergence the API-priority design handles.

**Polish points (judged with eyes on screen):**
- (b) Card placement: **kept below both status rows.** The KeyPath.app + kanata-launcher rows are a grouped pair; inserting the card between them splits the pair, and the card's own copy already names KeyPath. Grouping wins.
- (a) 1.5s auto-open-Settings vs 12s card: acceptable as-is. "Open the destination, then explain what to do there" is a defensible pattern; the mild step-1/button redundancy isn't worth churn in this PR. Noted as possible future polish (raising the wizard above Settings when the card appears).

No code changes needed.

Issue #896, opened 2026-06-11
RemapEndToEndTests fails on the self-hosted CI runner (simulator returns no mappings) — passes locally with identical engine


malpern:
Unmasked by [#891](https://github.com/malpern/KeyPath/pull/891)'s double-fulfill fix. Before that fix, the full coverage lane crashed mid-run and `KEYPATH_ALLOW_TEST_RUNNER_CRASH_SUCCESS=1` reported the lane green anyway — `RemapEndToEndTests` likely never executed on the runner. With the crash fixed, the lane runs further than ever and these two tests fail consistently:

```
testRuleCollectionRemapOneToTwo: XCTAssertEqual failed: ("nil") is not equal to ("Optional("2")")
testRuleCollectionRemapRoundTrip: Key a must output b / Key 1 must output 2 / Key q must output w — all nil
```

## Ruled out (verified during #891 babysitting)

- **Engine pin / stale simulator**: a poisoned workspace cache was real and is fixed (engine-sha stamp, commit on #891) — the failing run rebuilt fresh from the correct pin (`cache miss: cached=none expected=6a05a4a…`) and still failed.
- **Binary identity**: CI and `Scripts/build-kanata-simulator.sh` both deploy `kanata_simulated_input` (renamed `kanata-simulator`).
- **The product itself**: same engine passes locally (`KEYPATH_SIMULATOR_PATH=build/kanata-simulator swift test --filter RemapEndToEndTests` → green) and the installed-app smoke suite exercises real remaps end-to-end through real kanata.

## Remaining suspects

1. Full-lane test-order pollution (these pass in isolation; the lane runs ~4300 tests in one process).
2. Runner env: `CI_ENVIRONMENT=true`, `KEYPATH_BUNDLED_SIMULATOR_OVERRIDE=/opt/homebrew/bin/kanata-simulator`, sandbox/exec context for spawning the simulator from the test process.
3. The simulator's stdout/stderr on the runner (next diagnostic: instrument `SimulatorService` to dump raw output on parse-to-nil and re-run on CI).

## Interim state

The two tests are skipped when `CI_ENVIRONMENT=true` (same pattern as `SKIP_EVENT_TAP_TESTS`), with this issue referenced. Local runs and the installed-app smokes keep real coverage of the remap pipeline.

## Related cleanup

[...]

PR #951, opened 2026-07-04
ci: reset ~/.config/keypath between runs on self-hosted runner

malpern:
## Problem

The self-hosted runner's `\$HOME` persists between runs. Pack/collection tests write to `~/.config/keypath` via `NSHomeDirectory()` ([KeyPathConstants.swift](Sources/KeyPathCore/KeyPathConstants.swift) — no config-dir env override), and `installed-packs.json` lives there too. A crashed or interrupted prior run can leave that directory in a state where enabling a rule collection fails at [PackInstaller.swift:132](Sources/KeyPathAppKit/Services/Packs/PackInstaller.swift) (`toggleCollection(isEnabled: true)` → `saveFailed("could not enable associated rule collection")`).

Because re-runs land on the **same** runner with the **same** `\$HOME`, this is a **deterministic wedge, not a flake**: once the state goes bad, every subsequent run fails until someone clears it by hand.

## Evidence

This blocked PR #945 across **three** runs — 5 `CLIPackCRUDTests` failures each time, all `could not enable associated rule collection` — even though #945's diff is **view-only** (`HoldTimingSliderRow.swift`, `PackDetailView.swift`) and cannot reach the `PackInstaller` path. `master` was green only because it last ran *before* the state went bad. Same persistent-state class as #896 / #891 (the kanata engine-sha stamp already fights the same "workspace persists" reality for the binary).

## Fix

Add a **Reset persistent test state** step before the test lane in the `build-and-test` job that clears `~/.config/keypath`, so every run starts hermetic. The suite recreates whatever it needs.

**Self-healing:** this PR's own CI run executes the new step first, clearing the currently-poisoned state — so it should go green on its own, and unblock #945 once #945 is rebased onto it.

## Scope

- Only the `build-and-test` self-hosted job needs it; `code-quality` only lints changed Swift files and never touches `~/.config/keypath`.
- Safe on a dedicated CI runner: `~/.config/keypath` there is purely test scratch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,6 +90,28 @@ jobs:
         chmod +x ./Scripts/lint-no-subprocess-in-autofixer.sh
         ./Scripts/lint-no-subprocess-in-autofixer.sh
 
+    - name: Reset persistent test state
+      # The self-hosted runner's $HOME persists between runs (same lesson as the
+      # kanata engine-sha stamp above). Tests that install packs or toggle rule
+      # collections write to ~/.config/keypath via NSHomeDirectory() — there is
+      # no config-dir env override. A crashed or interrupted prior run can leave
+      # that directory in a state where enabling a collection fails, which then
+      # breaks CLIPackCRUDTests (and any other pack/collection test) on *every*
+      # subsequent run on the same runner until it is cleared by hand — a
+      # deterministic wedge, not a flake. This was the root cause of the #945 CI
+      # block (5 CLIPackCRUDTests failures: "could not enable associated rule
+      # collection"), the same persistent-state class as #896/#891. Reset to a
+      # clean slate so each run is hermetic; the suite recreates what it needs.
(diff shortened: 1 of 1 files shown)


malpern:
Closing in favor of #952. My diagnosis here was wrong: this PR's own CI run executed the reset step, confirmed it deleted `~/.config/keypath` (keypath.kbd, RuleCollections.json, installed-packs.json, .backups, pack-snapshots), and CLIPackCRUDTests **still failed identically** — so persistent home-dir state was never the cause.

The real cause is concurrency, correctly identified and fixed in #952: multiple PRs' `build-and-test` jobs run in parallel on the single self-hosted runner and race `RuleCollectionsManager`'s shared config file, producing the spurious `could not enable associated rule collection` failures. `fix/stuck-splash-on-reopen` failed in the same overlap window. Serializing the runner with a `concurrency:` group is the right fix; nuking state per-job is not.

PR #969, opened 2026-07-04
Fix kanata IM detection dead-end: functional evidence + continue path (#931)

malpern:
## Summary

Part of #931 (not auto-closed: on-device verification remains — the human-in-loop half of the ticket).

The KeyPath.app half of #931 was fixed by #937/#939 (authoritative `IOHIDCheckAccess` + manual-fallback escalation). The **kanata-launcher row still had the identical dead-end one level down**: its Input Monitoring state came only from the TCC database, so on macOS 26/27 — where a completed grant may never produce a readable TCC row — the wizard showed "Not yet added" forever, with no forward navigation on the page.

### Oracle: functional evidence for kanata IM

- New injectable `KanataFunctionalEvidence` provider on `PermissionOracle` (#931). A running, TCP-responding kanata with no diagnosed input-capture failure is strong evidence of the grant, so it upgrades an **absent** TCC read to `granted`.
- **Explicit TCC answers are authoritative in both directions** (revised per review): the underlying capture signal is optimistic-by-default upstream, so evidence never overrides an explicit `granted` or `denied` row — a mid-run revoke shows `denied` correctly. Evidence is only consulted when the TCC read is inconclusive (`nil`), the exact grant-without-a-row case this exists for.
- Negative or absent evidence never downgrades anything (a stopped daemon proves nothing).
- Wired from `configureWizardDependencies` via `ServiceHealthChecker`'s TTL-cached runtime snapshot — cheap under the wizard's 1s poll, and the `KeyPathPermissions` module gains no new dependency (closure injection only).
- Pure static resolver `resolveKanataInputMonitoring(tccStatus:evidence:)` mirrors the #937 pattern for unit testing.

### Wizard: the kanata row gets a forward path

- The kanata row now escalates like the KeyPath row (#939): when `unknown` persists past a 30s window (clock restarts when the drag-to-authorize flow is launched), a card offers **"Continue Setup"** — it navigates to the next wizard step (typically the Service page; deliberately **not** an auto-start, preserving the one-permission-prompt-at-a-time flow). Once the user starts the engine there, the Oracle upgrades on functional evidence; a genuinely missing grant fails with an input-capture error that `SystemInspector` routes straight back to this page. The loop is closed; the page can no longer strand the user.
- `denied` never reaches the bypass at either layer: the escalation card is gated on `.warning` (unknown), and the Oracle keeps an explicit denied row denied.
- Clock logic extracted to pure `advanceUnverifiedClock` (KeyPathWizardCore) with unit tests (start-once, no refresh while unknown, clear on conclusive, restart on re-entry).
- New accessibility id: `wizard_input_monitoring_kanata_continue_unverified`.

### Tests

- `PermissionOracleInputMonitoringTests`: truth table for the resolver — evidence upgrade from absent TCC, explicit granted/denied never overridden, stopped/half-started daemon never downgrades, diagnosed capture failure is not proof, nil-provider passthrough identical to pre-#931.
- `UnverifiedClockTests`: the four clock transitions.

## Validation

- **Authored on Linux — CI is the compile gate.** All API signatures verified against the codebase by reading.
- The `/thermo-nuclear-swift-review` gate should run against this branch before merge (unavailable in the authoring environment). Two adversarial review passes already ran; findings fixed in `2e94dfb` (denied-override race, clock test coverage) and `0e95578` (continue-card copy matched to actual flow).
- Known accepted looseness: provider registration happens in an un-awaited `Task` at startup — a snapshot taken in that window just sees no evidence (conservative), and the wizard's 1s poll self-heals. `functional-im` results report `.high` confidence, which is inert today (logging only) — if `Confidence` ever gates behavior, introduce a distinct tier.
- On-device test plan (macOS 26/27, fresh install): reach the IM page with kanata-launcher granted-but-rowless → after 30s the continue card appears → Continue Setup → Service page → click Start → IM row flips green via `kanata.functional-im` source (check log), wizard completes. Negative paths: skip the grant → engine start fails with input-capture error → routed back with error severity; revoke the grant while running → row shows denied (TCC), no false green.

https://claude.ai/code/session_0142A1G18McVBb1f1Dp7873M

### DIFF
--- a/Sources/KeyPathAppKit/WizardProtocolConformances.swift
+++ b/Sources/KeyPathAppKit/WizardProtocolConformances.swift
@@ -198,6 +198,21 @@ public func configureWizardDependencies(runtimeCoordinator: RuntimeCoordinator) 
             onAutoFix: onAutoFix
         ))
     }
+
+    // Functional evidence for kanata's Input Monitoring resolution (#931):
+    // on macOS 26/27 a grant may never produce a readable TCC row, so the
+    // Oracle falls back to runtime proof. ServiceHealthChecker's runtime
+    // snapshot is TTL-cached, keeping this cheap under the wizard's 1s poll.
+    Task {
+        await PermissionOracle.shared.setKanataFunctionalEvidenceProvider {
+            let snapshot = await ServiceHealthChecker.shared.checkKanataServiceRuntimeSnapshot()
+            return PermissionOracle.KanataFunctionalEvidence(
+                isRunning: snapshot.isRunning,
+                isResponding: snapshot.isResponding,
+                inputCaptureReady: snapshot.inputCaptureReady
+            )
+        }
+    }
 }
 
 /// Configure WizardDependencies for app-bundled CLI repair/inspect entry points.
--- a/Sources/KeyPathInstallationWizard/UI/Pages/WizardInputMonitoringPage.swift
+++ b/Sources/KeyPathInstallationWizard/UI/Pages/WizardInputMonitoringPage.swift
@@ -20,6 +20,12 @@ public struct WizardInputMonitoringPage: View {
     /// triggered. Drives the escalation to manual guidance when the grant never
     /// registers (macOS 26/27 dead-end, #931).
     @State private var keyPathRequestAttemptedAt: Date?
+    /// When kanata-launcher's Input Monitoring was first observed `.unknown` on
+    /// this page (or when the user last launched the add flow, whichever is
+    /// later). On macOS 26/27 a completed grant may never produce a readable
+    /// TCC row, so after a wait window we must offer a way FORWARD — starting
+    /// the engine is the only reliable verification left (#931).
+    @State private var kanataUnverifiedSince: Date?
 
     @Environment(WizardStateMachine.self) private var stateMachine
 
@@ -203,6 +209,10 @@ public struct WizardInputMonitoringPage: View {
                                                 return
                                             }
                                             AppLogger.shared.log("🔧 [WizardInputMonitoringPage] Kanata Fix clicked — presenting drag-to-authorize overlay")
+                                            // Restart the escalation clock: the user is
+                                            // mid-flow, so give the full wait window
+                                            // before offering the continue path (#931).
+                                            kanataUnverifiedSince = Date()
                                             DragToAuthorizeController.shared.present(for: .inputMonitoring)
                                         }
                                     }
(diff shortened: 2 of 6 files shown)


malpern:
Review findings addressed in `2e94dfb`:

- **Denied-override race (main finding): fixed.** Confirmed the trace — `inputCaptureReady` is optimistic-by-default upstream (`.ready` unless a failure was affirmatively diagnosed), so it must not outrank an explicit TCC answer. `resolveKanataInputMonitoring` now returns any explicit TCC result untouched; evidence only upgrades an absent (`nil`) read, and the provider is only consulted in that case. `testFunctionalEvidenceOverridesStaleDeniedRow` replaced with `testExplicitTCCAnswerIsNeverOverriddenByEvidence` covering both directions. Doc comments and the PR description no longer claim "kernel-level proof."
- **Clock test coverage: added.** `updateKanataUnverifiedClock` now delegates to a pure `advanceUnverifiedClock(previous:statusIsUnknown:now:)` in KeyPathWizardCore, with tests for start-once, no-refresh-while-unknown (intentional, now documented), clear-on-conclusive, and restart-on-re-entry.
- **Startup registration race: consciously accepted.** A snapshot taken before the un-awaited registration completes just sees no evidence and stays conservative; the 1s poll self-heals. Noted in the PR description.
- **Grab-store-positive proof:** not adopted for now — requiring `KanataGrabStatusStore.latest?.active == true` would make the upgrade unreliable wherever `InputGrab` messages don't arrive, and with the denied-override closed, the residual optimistic window only affects the absent-row case where a genuinely missing grant surfaces as a diagnosed capture failure within one health-cache TTL.

---

malpern:
Confirmed and fixed in `0e95578` — the Service page indeed only refreshes on appear; the user clicks Start. Went with the copy-softening option rather than auto-start: kanata's Accessibility may also be ungranted at this point in a fresh install, and auto-starting from the IM page would fire both system prompts at once — the exact thing `routeForUnverifiedKanataPermissions` exists to prevent. Card now says "you'll start the engine on the next step," the struct doc records the deliberate no-auto-start decision, and the PR test plan is corrected to "Continue Setup → Service page → Start."

On the `.high` confidence flag for `functional-im`: agreed it's cosmetically wrong for a heuristic signal, and agreed it's inert today (logging only). Leaving as-is to keep this PR scoped; if `Confidence` ever gates behavior, that refactor should introduce a distinct tier rather than reuse `.low`.

---

malpern:
Closing as stale draft. The wizard/input-monitoring flow has changed substantially since this branch, and it now conflicts with master. Re-open as a fresh PR if the macOS rowless-TCC functional-evidence problem still reproduces.

PR #1083, opened 2026-07-10
Verify installer recipe postconditions

malpern:
## Summary
- declare operation-specific observable postconditions on installer recipes
- verify those declarations against the engine-owned fresh final snapshot
- treat satisfied final state as authoritative after a lost operation reply, and attach a newly planned repair when verification fails
- record structured verification telemetry and update the Milestone 3 progress log

## Validation
- `TEST_FILTER=InstallerEngineEndToEndTests ./Scripts/run-tests-safe.sh` (14 passed)
- `TEST_FILTER=InstallerEnginePlanTests ./Scripts/run-tests-safe.sh` (3 passed)
- `TEST_FILTER=InstallerEngine ./Scripts/run-tests-safe.sh` (106 passed)
- `./Scripts/run-tests-safe.sh` (full non-snapshot suite passed; 1 machine-integration test skipped)
- `git diff --check`

## Review gate
- Local gate selected the enforced remote review path (exit 2); do not merge until `claude-review` passes.


malpern:
Addressed the review finding in 7bb3cbdd. Verified-success recovery now requires the specific failed recipe to declare non-empty postconditions and for those postconditions to hold in the final snapshot. Added a multi-recipe regression test proving an unrelated recipe postcondition cannot mask a failure; the focused end-to-end suite passes 15/15.

malpern:
Addressed the pre-existing-state masking findings in 9367b2a9. Generated and single-action plans now retain postcondition satisfaction from the complete pre-run snapshot. A failed operation is converted to verified success only when every postcondition owned by that recipe transitions from false before execution to true afterward; a missing/incomplete baseline remains a failure. Recovery planning is now limited to install/repair intents. Added regression coverage for already-satisfied state, baseline capture, and reran 109 installer tests plus the full local suite successfully.

malpern:
Addressed the final observability notes in 8592f900. Verification now deduplicates postconditions only from the executed recipe prefix, so skipped recipes cannot emit misleading satisfaction telemetry. Added a regression test proving a non-empty executed contract fails conservatively when no final snapshot is available. Focused installer suite: 110 passed; prior full suite remains green.

malpern:
Addressed both final review notes in ac5d6195. `terminateConflictingProcesses` is now explicitly a mutating `resolveRequirement` recipe. Each `RecipeResult` carries the postconditions declared by the recipe that actually ran, and verification derives directly from those outcomes rather than plan position. Installer suite remains green at 110 tests.

malpern:
The conflict-scope question exposed a real mismatch, fixed in 7490ee57. Conflict-resolution recipes now carry the exact conflicts captured by planning. Execution resolves external Kanata conflicts via the Kanata process operation, disables the Karabiner grabber for that conflict class, deduplicates each operation, and fails explicitly for unsupported conflict types instead of claiming global success. Added successful mixed-conflict routing and unsupported-conflict tests; routing 7/7 and end-to-end 17/17 pass.

malpern:
The CI lifecycle failures were causally related and are fixed in e6e1efe1. Planned repairs still resolve their exact captured conflict set, while a trusted direct stale-runtime request with no general conflict entry preserves the historical explicit Kanata termination fallback. Exact validation: ServiceLifecycleCoordinatorTests 23/23 and InstallerEngineSingleActionRoutingTests 7/7.

malpern:
Addressed the order-dependence in 91f7855e. Conflict execution now accumulates unsupported entries, resolves all supported conflicts first, then reports the unsupported remainder. Regression covers unsupported-first ordering while confirming Kanata termination still runs. Routing 7/7 and lifecycle 23/23 pass.

malpern:
Addressed shared-postcondition attribution in 4973a88f. Lost-reply verified success is now disallowed when any earlier executed recipe declared the same postcondition, preventing an earlier coarse runtime transition from excusing a later operation failure. Added the two-recipe shared-contract regression; end-to-end 18/18 and routing 7/7 pass.

malpern:
Confirmed both final points as intended invariants. Plans built by `run` and `runSingleAction` execute immediately inside the transaction; if an externally reused plan goes stale, the fresh final snapshot must fail verification and produce a newly classified recovery plan rather than letting the executor make a live planning decision. Execution always breaks on first failure, and `InstallerEnginePlanTests.testExecuteSkipsRecipesAfterFailure` already pins this with a three-recipe plan and asserts only the first recipe executes.

PR #1084, opened 2026-07-10
Correlate installer run evidence

malpern:
## Summary
- assign stable IDs to snapshots, plans, and installer runs
- correlate before/after evidence through reports and per-step telemetry
- expose shared structured completion and recovery states through app and CLI report paths
- preserve approval-pending as a successful operation that still requires user action

## Validation
- `TEST_FILTER=InstallerEngine ./Scripts/run-tests-safe.sh` (114 passed)
- `TEST_FILTER=CLIReportModelsTests ./Scripts/run-tests-safe.sh` (4 passed)
- `TEST_FILTER=CLIOutputContractTests ./Scripts/run-tests-safe.sh` (12 passed)
- broad non-visual suite: 3,690 passed, 41 skipped

## Gate note
- the separate Xcode-beta snapshot target hit its existing renderer/harness crash (`CGRectValue` exceptions and signal 11); this slice does not change UI rendering

### DIFF
--- a/Sources/KeyPathAppKit/CLI/SystemFacade.swift
+++ b/Sources/KeyPathAppKit/CLI/SystemFacade.swift
@@ -504,12 +504,15 @@ public extension CLIInstallerReport {
         let finalIssues = SystemFacade.issues(from: finalContext ?? initialContext)
         let finalUserActionIssues = finalIssues.filter { !$0.canAutoFix }
         let finalOperational = finalContext.map(SystemFacade.isOperational) ?? false
-        let completedButStillBlocked = report.success && !finalOperational && !finalIssues.isEmpty
+        let completedButStillBlocked = report.success
+            && report.completionState != .awaitingApproval
+            && !finalOperational
+            && !finalIssues.isEmpty
         let failedForManualApproval = report.failureReason.map(SystemFacade.requiresManualApproval) ?? false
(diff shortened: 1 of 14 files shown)


malpern:
Addressed both review points in 597a8975: the unconfigured-uninstall early return no longer captures system state (with a regression test proving zero validator reads), and normal CLI reports preserve the prior tri-state wire contract by omitting `userActionRequired` unless it is true. Configured uninstall still captures before/after evidence as required by the owned-run contract.

malpern:
Addressed the re-review consistency points in f15c59dc: configured uninstall now forces fresh captures for both before and after evidence, with a test pinning both IDs, both `.fresh` reads, and telemetry correlation; blocked dry-run CLI plans now emit `completionState: blocked`.

malpern:
Keeping the two fresh captures intentionally. Milestone 3’s decision model is explicitly `fresh before -> execute -> fresh after`; uninstall is a low-frequency destructive workflow where trustworthy evidence outweighs cached-read latency, while the unconfigured early-return remains probe-free. Also intentionally prioritizing `awaiting-approval` as the actionable terminal state when it overlaps a verified lost reply; the underlying operation error remains preserved in `executedRecipes` and repair telemetry rather than being discarded.

PR #1085, opened 2026-07-10
Execute only declared installer actions

malpern:
## Summary
- plan VHID activation from captured device evidence and order it before dependent service work
- remove executor-side VHID probes and injected activation branches from generic install/restart recipes
- add a lint ratchet so generic recipe execution cannot regain planning decisions
- require observable final postconditions for every mutating recipe, with three named non-runtime/no-op exemptions

## Validation
- `TEST_FILTER=InstallerEngine ./Scripts/run-tests-safe.sh` (117 passed)
- `TEST_FILTER=InstallerDecisionPipelineLintTests ./Scripts/run-tests-safe.sh` (5 passed)
- `TEST_FILTER=PostconditionLintTests ./Scripts/run-tests-safe.sh` (6 passed)

## Gate note

### DIFF
--- a/Sources/KeyPathInstallationWizard/Core/ActionDeterminer.swift
+++ b/Sources/KeyPathInstallationWizard/Core/ActionDeterminer.swift
@@ -113,7 +113,10 @@ public enum InstallerDecisionPipeline {
         // Check if daemon needs starting
         if !context.services.karabinerDaemonRunning {
             // Ensure manager is activated before starting daemon
-            if !context.components.vhidDeviceInstalled, !actions.contains(.activateVHIDDeviceManager) {
+            if !context.components.vhidDeviceInstalled,
+               !context.requiresManualVHIDDriverApproval,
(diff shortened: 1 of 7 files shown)


malpern:
The terminal manual-approval return already prevented the reported path, but I made the gate local to both daemon-start guards in 7938f4c6 and added a table test proving neither install nor repair plans activation/daemon start while approval is pending.

malpern:
Correction: the follow-up commit is c5b5c3a9.

malpern:
Confirmed both points. The full function has a terminal `if context.requiresManualVHIDDriverApproval { return actions }` before the daemon-start block; the new test executes the full decision function for both intents and proves neither activation nor daemon start is planned. The field change is intentional: driver installation is payload presence, while `vhidDeviceInstalled` is the captured evidence that manager activation produced the device; when the driver is absent both are false, so ordering remains install payload -> activate manager -> start services.

PR #1086, opened 2026-07-10
Render wizard from owned run result

malpern:
## Summary
- run full and single-action wizard fixes through the owned InstallerEngine run API
- project the engine-verified final context directly into wizard state
- render structured completion states, including approval-pending, without client-side reclassification
- remove delayed post-fix refreshes, SMAppService preflight reads, detached VHID probes, and dead diagnostic helpers
- add a lint ratchet preventing second observers from returning to wizard action handling

## Validation
- `TEST_FILTER=InstallerDecisionPipelineLintTests ./Scripts/run-tests-safe.sh` (6 passed)
- `TEST_FILTER=WizardStateRegressionTests ./Scripts/run-tests-safe.sh` (2 passed)
- `TEST_FILTER=WizardStateMachineNavigationTests ./Scripts/run-tests-safe.sh` (27 passed)
- `./Scripts/test-fast.sh ui` (327 passed)

## Gate note

### DIFF
--- a/Sources/KeyPathInstallationWizard/UI/InstallationWizardView+Actions.swift
+++ b/Sources/KeyPathInstallationWizard/UI/InstallationWizardView+Actions.swift
@@ -22,11 +22,16 @@ public extension InstallationWizardView {
             defer { Task { @MainActor in fixInFlight = false } }
 
             let report = await InstallerEngine().run(intent: .repair, using: PrivilegeBroker())
-            applyOwnedRunResult(report)
+            let evidenceApplied = applyOwnedRunResult(report)
 
             // Show result toast
             await MainActor.run {
-                if report.completionState == .awaitingApproval {
+                if report.success, !evidenceApplied {
+                    toastManager.showError(
+                        "Repair finished, but KeyPath could not verify the final system state. Close and reopen the wizard before retrying.",
+                        duration: 8.0
+                    )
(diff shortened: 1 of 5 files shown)


malpern:
Addressed the missing-evidence case in the latest commit: a successful report without `finalContext` now produces a user-visible verification error and returns false rather than showing success or advancing. Failed/blocked reports continue to show their structured failure reason. I also verified there are no `.wizardStartupRevalidate` observers anywhere in Sources; the only remaining use is another post on wizard dismissal. The app observes `.kp_startupRevalidate`, which is a distinct notification with a different raw value.

malpern:
Confirmed: this builds cleanly because `InstallationWizardView` is main-actor isolated under the current SwiftUI toolchain; adding `await` produced the compiler warning “no async operations occur within await,” so the synchronous calls are correct (the surrounding `MainActor.run` blocks are pre-existing/redundant cleanup for Milestone 4). As noted above, `.wizardStartupRevalidate` has no observer and its raw name differs from the app’s `.kp_startupRevalidate`. The success-to-failure behavior without final evidence is intentional fail-closed policy: all mutating ready-plan paths capture final context even on operation failure; nil is limited to non-mutating blocked/inspect paths or a contract regression, and callers must not advance on an unverified mutation.

PR #1094, opened 2026-07-10
Verify postconditions before privileged fallback

malpern:
## Summary
- generalize postcondition-before-fallback handling across helper-first runtime recovery, newsyslog install, VHID driver install, process termination, and Kanata shutdown
- recheck uninstall postconditions after a failed helper reply before Emergency Cleanup
- add per-operation regression coverage and move newsyslog out of the postcondition exemption list
- update the helper race bug record with the generalized invariant

## Validation
- `TEST_FILTER='PrivilegedOperationsRouterTests|UninstallCoordinatorTests|PostconditionLintTests' ./Scripts/run-tests-safe.sh` (50 passed)
- `TEST_FILTER=KarabinerConflictServiceTests ./Scripts/run-tests-safe.sh` (4 passed)
- `git diff --check`

## Broad gate note
`KEYPATH_TEST_RESET_MODULE_CACHE=1 ./Scripts/test-fast.sh installer` passes 613 selected tests and fails only `VirtualHID daemon detection uses injected SystemStateProvider`. The exact same failure reproduces from untouched `origin/master` in a separate clean worktree, while the conflict-service suite passes in isolation. This is baseline global-state interference, not introduced by this diff.

## Review gate
Local review tooling selected the remote review gate; do not merge until `claude-review` passes.

### DIFF
--- a/Sources/KeyPathAppKit/Core/PrivilegedOperationsRouter.swift
+++ b/Sources/KeyPathAppKit/Core/PrivilegedOperationsRouter.swift
@@ -40,6 +40,17 @@ public final class PrivilegedOperationsRouter {
         nonisolated(unsafe) static var installAllServicesOverride: (() async throws -> Void)?
         nonisolated(unsafe) static var recoverRequiredRuntimeServicesOverride: (() async throws -> Void)?
         nonisolated(unsafe) static var killExistingKanataProcessesOverride: (() async throws -> Void)?
+        nonisolated(unsafe) static var helperRecoverRequiredRuntimeServicesOverride: (() async throws -> Void)?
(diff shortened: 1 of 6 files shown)


malpern:
Addressed both review points in d45d1f17:

- Removed the early return so helper-error recovery still performs post-restart service discovery and the final runtime postcondition check; the regression test now requires two readiness checks.
- Limited the extra uninstall postcondition recheck to actual failed helper replies, preventing completed-helper fallthrough from re-running unregister/verify steps. Added a unique-step-ID Emergency Cleanup regression test.

Focused gate: 51 passed.

PR #1099, opened 2026-07-10
Harden helper timeout recovery

malpern:
## Summary
- make queued helper gate cancellation safe and normalize interrupted XPC replies as ambiguous outcomes
- guard connection cleanup by generation so late callbacks cannot clear a replacement
- give ambiguous mutations a bounded recovery window and propagate incomplete helper evidence into the canonical snapshot
- document the incident and add focused regression coverage

## Validation
- stable Xcode 26.6 KeyPath product build passed
- focused safe gate passed: 55 tests
- accessibility check passed
- full safe suite: 1,932 passed; 24 pre-existing AppKit snapshot harness failures and xctest signal 11, tracked as the approved snapshot-harness follow-up

## Review gate
Local gate selected the enforced remote claude-review path (exit 2). Do not merge until that check passes.

### DIFF
--- a/Sources/KeyPathAppKit/Core/HelperManager+ConnectionLifecycle.swift
+++ b/Sources/KeyPathAppKit/Core/HelperManager+ConnectionLifecycle.swift
@@ -46,15 +46,19 @@ extension HelperManager {
         // Set up the interface
         newConnection.remoteObjectInterface = NSXPCInterface(with: HelperProtocol.self)
 
-        // Handle connection lifecycle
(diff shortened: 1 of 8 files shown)


malpern:
Addressed both review notes: added a direct stale-generation callback regression test, and moved the bounded recovery delay before operation-gate acquisition so health probing cannot hold the serialized permit while it waits. Focused HelperManager and SystemValidator gate passes (24 tests).

PR #1101, opened 2026-07-10
Capture TCP configuration in system snapshot

malpern:
SystemValidator now captures daemon TCP configuration as canonical HealthStatus evidence, validates a real numeric port, and MainAppStateController derives readiness from that snapshot instead of reparsing the plist. Includes health and parser tests, a structural lint ratchet, and durable bug documentation. Validation: stable Xcode 26.6 build, 159 consumer tests, 25 focused snapshot tests, accessibility pass. Remote claude-review is required before merge.

### DIFF
--- a/Sources/KeyPathAppKit/Services/MainAppStateController.swift
+++ b/Sources/KeyPathAppKit/Services/MainAppStateController.swift
@@ -268,62 +268,6 @@ class MainAppStateController {
         )
     }
 
-    // MARK: - TCP Configuration Check
-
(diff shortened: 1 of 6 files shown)


malpern:
Confirmed the review point: lastTCPConfigured is optional for initial UI state. I made the post-capture assignment explicitly fail-safe with nil coalescing to false, preserving the old unproven-means-misconfigured behavior while keeping raw optional evidence in HealthStatus. Focused controller and lint tests pass.

PR #1109, opened 2026-07-11
Fix stale uninstaller resource paths

malpern:
## Summary

- repair the `Scripts/uninstall.sh` relative symlink after the app source move
- update the development uninstaller fallback and build warning to `Sources/KeyPathApp/Resources`
- add a deployment contract test covering the symlink target, executable resolution, coordinator fallback, and warning text

## Root cause

The app resources moved from `Sources/KeyPath` to `Sources/KeyPathApp`, but three uninstaller references were left behind. The dangling script symlink caused repository rsync traversal to fail before CrabBox could run.

## Validation

- `TEST_FILTER=DeploymentScriptContractTests ./Scripts/run-tests-safe.sh` — passed, 5 XCTest cases
- `./Scripts/test-full.sh` — passed, 4,733 tests
- `mise exec -- swiftformat --lint Tests/KeyPathTests/Lint/DeploymentScriptContractTests.swift Sources/KeyPathAppKit/Managers/UninstallCoordinator.swift` — passed
- `python3 Scripts/check-accessibility.py` — passed
- `bash -n Scripts/build-and-sign.sh` — passed
- repository `rsync -a --dry-run` traversal excluding `.git` and `.build` — passed
- symlink resolves to an executable file — verified
- `git diff --check` — passed
- `./Scripts/review-gate.sh` — remote review gate selected; GitHub `claude-review` must pass before merge

## Installer matrix

Affected row: uninstall completion and failure. This patch restores resource discovery only; it does not change uninstall behavior or lifecycle semantics.

### DIFF
--- a/Scripts/build-and-sign.sh
+++ b/Scripts/build-and-sign.sh
@@ -394,7 +394,7 @@ if [ -d "Sources/KeyPathApp/Resources" ]; then
     fi
     echo "✅ Copied app resources"
 else
-    echo "⚠️ WARNING: Sources/KeyPath/Resources directory not found"
+    echo "⚠️ WARNING: Sources/KeyPathApp/Resources directory not found"
 fi
 
 # Copy SwiftPM resource bundles (KeyPath_KeyPath.bundle, KeyPath_KeyPathAppKit.bundle, etc.)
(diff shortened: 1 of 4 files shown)


malpern:
Thanks. The repository-wide search was run across non-archived production code and active scripts/docs; the old path now appears only in negative regression assertions. The `PWD`-based development fallback behavior predates this path correction and is intentionally unchanged in this focused fix.

PR #1116, opened 2026-07-12
Fix CLI tri-mode permission reporting

malpern:
## Summary
- preserve granted, denied/error, and unknown permission states in CLI diagnostics
- report unverifiable permissions as non-blocking warnings instead of missing permissions
- keep CLI operational and installer user-action results consistent with tri-mode semantics
- add CLI contract coverage for all three states

## Validation
- `swift test --filter CLIOutputContractTests --jobs 4`
- `./Scripts/test-full.sh` (4,739 passed)
- `git diff --check`

## Review gate
- Local review tooling selected the remote review gate; do not merge until `claude-review` passes.

### DIFF
--- a/Sources/KeyPathAppKit/CLI/SystemFacade.swift
+++ b/Sources/KeyPathAppKit/CLI/SystemFacade.swift
@@ -421,6 +421,7 @@ public struct SystemFacade: Sendable {
             permission: "Input Monitoring",
             deniedAction: "Open KeyPath.app and grant Input Monitoring in System Settings > Privacy & Security > Input Monitoring",
             remediationURL: WizardSystemPaths.inputMonitoringSettings,
+            isBlocking: false,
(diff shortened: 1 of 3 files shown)


malpern:
Addressed in defb91cc. KeyPath's own Input Monitoring is intentionally soft for core remapping, per `PermissionOracle.isSystemReady`; the omission from `isOperational` was deliberate. The inconsistency was the CLI issue severity, which still treated that soft permission as blocking. The follow-up now:

- reports KeyPath-own Input Monitoring denial as a warning/non-user-action issue
- keeps Kanata permissions and KeyPath Accessibility blocking
- uses a typed `CLISystemIssueSeverity` enum
- adds a mixed-status regression test proving KeyPath-own IM denial remains operational

Targeted CLI contract tests pass (15/15).

PR #1123, opened 2026-07-12
Make Kanata config checks asynchronous

### LINKED ISSUE #873: [Tests] Refactor kanata --check test helper: pipe-deadlock + MainActor concerns
Captured during the PR #871 / #864 review cycle for the 1.0 release push. Deferred from those PRs because the fix is a pre-existing pattern across multiple test files, not unique to either PR.

## What's there today

Several integration test suites share a `validateWithKanata` helper that:

\`\`\`swift
@MainActor
private func validateWithKanata(_ config: String) async throws -> (isValid: Bool, errors: [String]) {
    let process = Process()
    // ... setup pipes ...
    try process.run()
    process.waitUntilExit()                                   // ← blocking, on @MainActor
    let data = pipe.fileHandleForReading.readDataToEndOfFile()  // ← reads after wait
    ...
}
\`\`\`

Files with this pattern:
- `Tests/KeyPathTests/Integration/ConfigValidationTests.swift`
- `Tests/KeyPathTests/Integration/PerOptionMatrixTests.swift` (#864)
- `Tests/KeyPathTests/Integration/StubDeflayerSafetyNetTests.swift` (#871)

## The two concerns

1. **Pipe deadlock risk.** If kanata writes enough output to fill the OS pipe buffer before exiting, the process blocks waiting for a reader — but `waitUntilExit()` is called first, before any read. For \`kanata --check\` against small configs this never triggers in practice, but it's a latent trap for any future call site that passes a larger config or verbose flags.

2. **MainActor blocking.** The helper is `@MainActor` and `process.waitUntilExit()` is a synchronous blocking call. Holds the main actor for the full kanata invocation. In a test suite this just costs latency; the anti-pattern still matters because we're encoding it for anyone who copies the pattern.

## Proposed fix

Extract a single \`KanataCheckHelper\` (or extend \`MatrixTestHelpers\`):

\`\`\`swift
enum KanataCheckHelper {
    /// Runs \`kanata --cfg <tempfile> --check\` against the given config and
    /// returns its (exitCode, stderr lines). Detaches from the calling actor,
    /// reads the pipe concurrently with waitUntilExit so no buffer deadlock.
    static func runCheck(_ config: String) async throws -> (exitCode: Int32, errors: [String])
}
\`\`\`

Migrate all three existing call sites. The shape is well-understood; the work is mechanical.

## Out of scope

- Replacing the pattern in production code (this is test-only).
- Restructuring how integration tests find the kanata binary (separate concern).

malpern:
## Summary
- replace three duplicated `kanata --check` subprocess helpers with one async test helper
- drain combined stdout/stderr while Kanata runs so pipe capacity cannot deadlock the child
- centralize binary discovery, temporary config handling, exit status, and error parsing
- add a regression that writes 1 MiB before exit to prove concurrent draining

## Test plan
- `TEST_FILTER="KanataCheckHelperTests|ConfigValidationTests|PerOptionMatrixTests|StubDeflayerSafetyNetTests" ./Scripts/run-tests-safe.sh` (29 parsed passes)
- `./Scripts/test-fast.sh --changed` (full safe fallback: 4,742 passes)
- SwiftFormat 0.61.1
- `git diff --check`
- review gate: remote review gate selected

Fixes #873

### DIFF
--- a/Tests/KeyPathTests/Integration/ConfigValidationTests.swift
+++ b/Tests/KeyPathTests/Integration/ConfigValidationTests.swift
@@ -21,54 +21,10 @@ final class ConfigValidationTests: XCTestCase {
         continueAfterFailure = true
     }
 
-    private func findKanataBinary() -> String? {
-        let candidates: [String] = [
-            ProcessInfo.processInfo.environment["KEYPATH_KANATA_PATH"],
-            URL(fileURLWithPath: #filePath)
-                .deletingLastPathComponent() // Integration
-                .deletingLastPathComponent() // KeyPathTests
-                .deletingLastPathComponent() // Tests
-                .deletingLastPathComponent() // repo root
-                .appendingPathComponent("External/kanata/target/aarch64-apple-darwin/release/kanata")
-                .path,
-            "/Applications/KeyPath.app/Contents/Library/KeyPath/Kanata Engine.app/Contents/MacOS/kanata",
-            "/opt/homebrew/bin/kanata"
-        ].compactMap { $0 }
-
-        return candidates.first { FileManager.default.isExecutableFile(atPath: $0) }
-    }
-
     @MainActor
     private func validateWithKanata(_ config: String) async throws -> (isValid: Bool, errors: [String]) {
-        guard let binary = findKanataBinary() else {
-            throw XCTSkip("Kanata binary not found — skipping CLI validation")
-        }
-
-        let tempFile = FileManager.default.temporaryDirectory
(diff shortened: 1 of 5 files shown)


malpern:
Addressed the output-ordering review point in the updated head: the helper now assigns one dedicated GCD reader ownership of `readDataToEndOfFile()` from process launch through EOF. A lock-protected rendezvous completes only after both the ordered byte stream and termination status arrive, so there are no competing readers or append-order races. Re-ran the 1 MiB regression and all three migrated suites: 29 parsed passes.

PR #1125, opened 2026-07-12
Harden VM lab runtime validation

malpern:
## Summary
- persist an owned provisioning manifest as soon as CrabBox reports a lease ID
- verify provider inventory before treating an absent provisional lease as cleaned
- support SecurityAgent's unnamed secure password field without exposing the secret in argv, logs, or the guest clipboard
- derive the protected submit button's current AX geometry, deliver a synthesized pointer click, and require the SecurityAgent sheet to close
- add a reusable two-Kanata-client VirtualHID readiness and reconnect scenario

## Why
The VirtualHID v8 migration needed isolated runtime testing. This closes the lab gaps found during that work: durable ownership during streamed provisioning, safe cleanup evidence, constrained secure-field input, and repeatable multi-client recovery coverage.

## Validation
- `./Scripts/test-fast.sh --changed` (earlier full safe suite: 4,743 tests passed; final rerun reached the 240-second suite timeout after 2,000+ passing tests and no failures)
- `./Scripts/review-gate.sh` (`remote review gate selected`)
- disposable macOS 15.7.7 v6 baseline: two clients ready, alive through daemon loss, and recovered after restart
- separate disposable macOS 15.7.7 v8 comparison using upstream Kanata `c7978d4` and driverkit PR branch `4b2bded`: same result
- fresh disposable macOS 15.7.7 SecurityAgent validation: role-based field assignment succeeded, fresh AX button geometry resolved the protected control, synthesized pointer delivery closed the sheet, the privileged proof command completed, and no guest secret temp file remained
- real CrabBox output matched the authoritative `leased cbx_…` record consumed by the streaming parser; fixtures also cover misleading earlier IDs, unterminated output, and a lease reported immediately before warmup failure
- CrabBox macOS image source confirms the guest intentionally uses `NOPASSWD:ALL`; secure input therefore verifies the actual SecurityAgent sheet postcondition rather than `sudo`

All disposable Tart leases used for this validation were destroyed.


malpern:
Addressed the cleanup correctness concern in `HEAD`: a failed stop is now treated as already absent only when a fresh provider inventory command succeeds and an exact extracted lease ID is absent. Free-form destroy-log text is no longer used as evidence. A regression fixture verifies misleading “not found” text cannot mark an inventoried lease clean. I also renamed the second-client PID and replaced fixed daemon-loss sleep with bounded readiness polling. The focused lab shell suite passes.

malpern:
Addressed lease-ID selection in `HEAD`: the streaming path now accepts only CrabBox’s canonical `leased cbx_…` record or a launcher’s exact one-line `cbx_…` result, rather than any token embedded in diagnostic output. The fixtures now emit a misleading earlier `cbx_` diagnostic before the authoritative record and verify the provisional/final manifest still targets the newly leased VM. Focused lab tests pass.

malpern:
Addressed the unterminated-output case in `HEAD`: both streamed warmup loops now process a final non-newline chunk. The launcher and desktop fixtures exercise authoritative lease output without a trailing newline, while the slow-warmup fixture retains a newline so it can verify provisional ownership before the producer exits. Focused lab tests pass.

malpern:
Final VM validation found and fixed a real integration issue: Peekaboo 3.9 accepted the text query but resolved `OK` to an unrelated QuickLook element. The SecurityAgent path now derives the named button center from fresh AX geometry, uses synthesized foreground coordinate delivery, and requires the secure sheet to close. A fresh macOS 15.7.7 Tart lease passed the supported helper end to end; the privileged proof command completed and no secret temp file remained. The latest review gap is also covered: a fixture now reports a canonical lease ID and then exits 9, asserting nonzero create plus `status=provisioning-failed` and `provision_result=9`. Actual CrabBox output from both fresh leases used the canonical `leased cbx_…` record with trailing fields; misleading earlier tokens and unterminated final records remain covered. Focused lab tests pass, the branch is 0 behind master, and all disposable leases are destroyed.

malpern:
Addressed the multi-authoritative-record concern in `HEAD`: every canonical `leased cbx_…` record now gets an owned provisional manifest, while the final authoritative record remains the active lease used for verification. This avoids both wrong-first selection and unowned retry resources. The fixture emits a stale authoritative lease, a misleading diagnostic token, and the final lease; it verifies the stale candidate is owned/cleanable and the final lease is selected. The focused lab suite passes. The dashboard `data-working` observation was from #1131/#1132 landing on master while this branch was being synchronized; those commits are now common ancestors and are not part of #1125’s focused diff.

malpern:
Reviewed the final audit points. Earlier authoritative candidates intentionally remain owned with `status=provisioning`, an expiry, and pending cleanup: capacity accounting keeps them reserved until `destroy`/expiry cleanup, which is safer than silently discarding a possibly real retry resource. The fixture explicitly proves the earlier candidate is cleanable. The relevant EXIT trap only calls `release_admission_lock`; the failure branch performs that call before clearing signal traps, so no operation cleanup is skipped. I added explicit user-facing mappings for exit 77 (SecurityAgent sheet did not close) and 78 (invalid button geometry). Focused tests pass after merging the latest master changes, and the branch is 0 behind.

malpern:
The remaining behavior is intentional and bounded. Superseded authoritative candidates are not hidden: each has an owned `provisioning` manifest, consumes admission capacity, appears in `list`, carries the same maximum two-hour expiry, and is reaped by the existing `cleanup` path (the fixture demonstrates explicit destroy). We avoid automatically stopping a superseded candidate inside the still-running warmup because its provider lifecycle may still be involved in that retry. The inventory fallback is also scoped to the lab’s local Tart/Parallels control plane after a stop failure; it requires a successful fresh list and exact lease-ID absence. A failed/incomplete inventory command remains `cleanup-failed`, so ambiguous evidence is not accepted. Required CI and Claude review are green.

malpern:
Confirmed the final lease-safety points against the merged behavior. Superseded candidates are deliberately not hidden or assumed dead: each has an owned manifest, `status=provisioning`, `cleanup_status=pending`, and `expires_epoch`; `list` surfaces it, capacity accounting reserves it, explicit `destroy` supports it, and `cleanup` reaps it at expiry. Inline destruction would be unsafe because a second authoritative record does not prove the earlier provider resource is disposable. For unknown provisional resources, a failed stop is marked complete only after a fresh inventory command succeeds and the exact lease ID is absent; stop-log wording is never accepted. This is the provider-level absence postcondition and avoids permanently wedging capacity on a lease that never became a provider resource. All required checks, including `claude-review`, are green and the branch is current.

PR #1126, opened 2026-07-12
Serialize shared VM provider admission

malpern:
## Summary
- serialize shared Tart/Parallels VM acquisition with an atomic host-side admission lock
- enforce configurable provider capacity (Tart 1, Parallels 2 by default) and return exit 75 with owning lease evidence when full
- recover dead creator locks while refusing live lock contention
- document multi-agent admission behavior

## Validation
- production fail-closed check against occupied Tart pool: exit 75, owner leases listed, no VM created

## Review gate
- remote review gate selected; GitHub `claude-review` must pass before merge

## State-matrix relevance
Harness infrastructure only; no KeyPath installer routing or product postconditions changed.

### DIFF
--- a/Scripts/lab/remote.sh
+++ b/Scripts/lab/remote.sh
@@ -28,6 +28,7 @@ LEASES="$STATE_ROOT/leases"
 ARTIFACTS="$STATE_ROOT/artifacts"
 LOGS="$STATE_ROOT/logs"
 OPERATIONS="$STATE_ROOT/operations"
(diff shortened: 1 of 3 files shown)


malpern:
Addressed both review concerns in `0f0f8039`:

- admission locks are now provider-specific, so Tart and Parallels can provision concurrently; a regression test holds the Tart lock while successfully creating a Parallels lease
- the default bounded wait is now five minutes (3000 × 0.1s), matching real VM provisioning better while remaining configurable
- docs now state both behaviors explicitly

Focused shell suite passes.

PR #1127, opened 2026-07-12
Fix VM admission lock interruption cleanup

malpern:
## Summary
- terminate VM provisioning after INT, TERM, or HUP while releasing the provider admission lock
- recover aged incomplete admission locks whose creator died before writing owner metadata
- cover signal exit/cleanup and incomplete-lock recovery in the lab shell suite

## Root cause
PR #1126 used the same non-terminating trap for `EXIT`, `INT`, `TERM`, and `HUP`. A real signal therefore released the lock but allowed provisioning to resume. A hard kill between lock-directory creation and owner-file creation could also leave an unrecoverable lock.

## Validation
- `./Scripts/test-fast.sh --changed` (full safe suite: 4,743 tests passed)
- `./Scripts/review-gate.sh` (`remote review gate selected`)

Follow-up to the actionable review feedback on merged PR #1126.

### DIFF
--- a/Scripts/lab/remote.sh
+++ b/Scripts/lab/remote.sh
@@ -29,6 +29,7 @@ ARTIFACTS="$STATE_ROOT/artifacts"
 LOGS="$STATE_ROOT/logs"
 OPERATIONS="$STATE_ROOT/operations"
 HELD_ADMISSION_LOCK=
+HELD_ADMISSION_OWNER=
 
 die() { print -u2 "keypath-lab(remote): $*"; exit 1; }
 now_epoch() { date +%s; }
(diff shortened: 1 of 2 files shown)


malpern:
Addressed the race in `2d0ae586`: new admission locks are now claimed by atomically hard-linking a fully written owner record into the fixed lock path, so contenders can never observe a newly created lock without metadata. Legacy directory locks remain readable/recoverable for transition safety. I also added the requested diagnostic to the stale-lock assertion. The focused lab shell suite passes.

malpern:
Addressed the pending-owner cleanup gap in `HEAD`: `acquire_admission_lock` now owns temporary INT/TERM/HUP handlers while it is contending, removes the prewritten owner record on interruption, and clears those handlers before returning. The busy-path diagnostic is now an explicit branch rather than an `&&`/`||` chain. A new regression test interrupts a process waiting behind a live lock and verifies exit 143 with no orphan owner record. The focused lab shell suite passes.

PR #1128, opened 2026-07-12
Clean interrupted VM admission waits

malpern:
## Summary
- remove prewritten VM admission owner records when a waiting process receives INT, TERM, or HUP
- preserve signal-derived exit codes while contending for a live provider lock
- make busy-lock owner diagnostics explicit for file and legacy-directory locks
- add a regression test for cancellation while waiting behind a live lock

## Root cause
The atomic claim added in #1127 writes its owner record before polling the fixed lock path. The holding-process traps are installed only after acquisition, so cancellation during the wait could leave the unlinked owner record behind.

## Validation
- #1127 full safe suite: 4,743 tests passed

Follow-up to the final review comment on #1127, which auto-merged before that comment became actionable.

### DIFF
--- a/Scripts/lab/remote.sh
+++ b/Scripts/lab/remote.sh
@@ -30,6 +30,7 @@ LOGS="$STATE_ROOT/logs"
 OPERATIONS="$STATE_ROOT/operations"
 HELD_ADMISSION_LOCK=
 HELD_ADMISSION_OWNER=
+PENDING_ADMISSION_OWNER=
 
 die() { print -u2 "keypath-lab(remote): $*"; exit 1; }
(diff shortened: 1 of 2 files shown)


malpern:
Addressed trap ownership in `HEAD`: `create_lease` now installs the unified EXIT/INT/TERM/HUP cleanup before calling `acquire_admission_lock`. `release_admission_lock` cleans both pending and held state, so acquisition no longer installs or clears caller traps, the pre-trap owner-record window is gone, and the same handler would release both states if the flow ever grows to hold one while acquiring another. The focused lab shell suite passes.

Issue #879, opened 2026-06-11
[Discussion] Should the bundled kanata be compiled WITHOUT the `cmd` feature? (compile-time vs config-time defense for root command execution)


malpern:
**DECISION (2026-06-11): Option B — compile `cmd` out of the bundled engine.** Discussed and agreed. The capability serves no shipping KeyPath feature (all actions are push-msg/TCP dispatched in-app); compiling it out is the only option that closes the user-writable-config → root-exec escalation that the config-layer work (#874/#877) structurally cannot reach. Reversible via one build flag if a real `(cmd …)` constituency ever materializes.

---

## Migration is clean — verified against the fork parser

The critical risk (does a cmd-less binary reject grandfathered configs?) does **not** materialize:

- `parser/src/cfg/defcfg.rs:865` parses `danger-enable-cmd` into a flag **unconditionally** — not behind `#[cfg(feature = "cmd")]`. A config carrying `danger-enable-cmd yes` (every legacy generated config does) **loads fine** on a cmd-less binary; the flag is set and simply never consulted.
- `parser/src/cfg/cmd.rs` `parse_cmd` bails at parse time with `#[cfg(not(feature = "cmd"))]` → only configs with an actual `(cmd …)` **action** fail, loudly, with "cmd is not enabled for this executable."

So: no config rewrite, no grandfathered-user breakage for header-only configs; only genuine hand-written `(cmd …)` users are affected, exactly as intended.

## No kanata fork change

`--features` is a KeyPath build-script flag; the `cmd` feature still exists in the kanata source. No submodule bump, no `keypath/bundled` branch work, none of the fork-safety constraints apply.

## Done so far

- `Scripts/build-kanata.sh`: production engine now builds `--features tcp_server` (was `cmd,tcp_server`), with a comment explaining why. The binary is built fresh (not committed), so the next Mac build/release produces the cmd-less engine.

## Remaining (tracked in the PR)

- **Mac-gated:** rebuild + re-sign/notarize the engine; verify a `(cmd …)` config now fails parse while a `danger-enable-cmd yes`-only config still loads.
- **App-side teardown of #874 machinery** — scope being confirmed (the Settings "Config Command Actions" toggle now promises a capability the binary can't deliver and must go; how far to unwind `KanataCommandActionsPolicy`/grandfathering/repair-enforcement is the open question).
[...]

PR #1134, opened 2026-07-12
Remove retired command actions preference

### LINKED ISSUE #909: Post-1.0: clean up stale ConfigCommandActionsEnabled UserDefaults key
#879 removed `KanataCommandActionsPolicy` and the `cmd` engine feature entirely. The UserDefaults key it used — `"KeyPath.Security.ConfigCommandActionsEnabled"` — persists for any user who ran a build that wrote it. Nothing reads it now, so it's benign residue.

Cleanup: a one-line removal during an app migration / first-launch pass:

```swift
UserDefaults.standard.removeObject(forKey: "KeyPath.Security.ConfigCommandActionsEnabled")
```

Low priority, non-blocking for 1.0. Flagged in #879 review so the debt is visible and a future reader who greps for the key (and finds no consumer) understands why it's still in defaults.

malpern:
## Summary
- clear the retired `KeyPath.Security.ConfigCommandActionsEnabled` default when preferences load
- add regression coverage proving older persisted values are removed

## Why
#879 removed command actions and their policy, leaving this preference as benign but misleading residue on upgraded installs. This keeps the cleanup in the existing preference migration path without reintroducing a consumer for the retired key.

Closes #909.

## Validation
- `TEST_FILTER=PreferencesServicePersistenceTests/testRetiredCommandActionsPreferenceIsRemovedOnLoad ./Scripts/run-tests-safe.sh`
- `./Scripts/test-fast.sh --changed` — 4,744 tests passed
- `./Scripts/review-gate.sh` — remote review gate selected; GitHub `claude-review` required before merge

### DIFF
--- a/Sources/KeyPathAppKit/Services/Configuration/PreferencesService.swift
+++ b/Sources/KeyPathAppKit/Services/Configuration/PreferencesService.swift
@@ -453,6 +453,10 @@ final class PreferencesService: @unchecked Sendable {
     // MARK: - Initialization
 
     init() {
+        // The command-action feature was removed in #879. Clear its retired
+        // preference so older installs do not retain a misleading security flag.
+        UserDefaults.standard.removeObject(forKey: "KeyPath.Security.ConfigCommandActionsEnabled")
+
         // Load stored preferences or use defaults
         let protocolString =
             UserDefaults.standard.string(forKey: Keys.communicationProtocol)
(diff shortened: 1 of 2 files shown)


malpern:
Moved the regression into its own `Retired Preference Cleanup` test section. The focused test still passes.

malpern:
Confirmed: this identifier is permanently retired with the removed command-action security policy and must not be reused. The unconditional removal is intentional and idempotent; adding a migration-completion default would replace one stale key with another for a two-line cleanup. No change needed.

PR #1164, opened 2026-07-13
Fix Tart protected-click coordinate mapping

malpern:
## Summary
- deliver fresh Accessibility points directly to the Tart VNC viewport
- keep raw framebuffer coordinates explicit for diagnostics
- align shell coverage and lab documentation with the verified mapping

## Validation
- `zsh -n Scripts/lab/keypath-lab Scripts/lab/remote.sh`
- `./Scripts/review-gate.sh` (remote review gate selected)


malpern:
Closing before merge: the live VM showed that the apparent no-op was an authorization sheet, not sufficient evidence that the coordinate mapper itself is wrong. I will reopen a narrower fix only after an isolated postcondition proof.

Issue #919, opened 2026-06-12
Regression-test permissions, daemon, and TCC behavior on macOS 27 betas


malpern:
**Post-1.0 tracking, but should happen during the macOS 27 beta cycle (public release expected September 2026).** From the WWDC 2026 review — see `docs/wwdc-2026-macos-27-opportunities.md` (PR #911), §10.

## Why

Nothing announced at WWDC26 changes SMAppService, XPC, or the TCC/Input Monitoring surface on paper — but TCC behavior shifts are historically where macOS major releases bite this app (PermissionOracle exists for exactly this reason).

## Checklist (on each significant beta seed)

- [ ] `PermissionOracle` snapshots: IOHIDCheckAccess vs TCC DB precedence still correct (ADR-001, ADR-006)
- [ ] Accessibility + Input Monitoring grant/revoke flows and the installer wizard
- [ ] SMAppService daemon registration/approval flow (`KanataDaemonService`), including the Login Items UI changes in the new System Settings
- [ ] XPC helper code-signature validation still passes under the new SDK
- [ ] Sparkle update + notarization/Gatekeeper assessment on macOS 27 (`Scripts/verify-installed-app.sh`)
- [ ] Kanata daemon keyboard capture end-to-end (the macOS 27 design refresh touches system chrome; verify overlay windows render correctly)

https://claude.ai/code/session_01Nsiqm39oCwHkbrHytefnGM

malpern:
macOS 27 beta checkpoint — 2026-07-11

Merged PR #1115 adds the repeatable strict capture and disposable-lab scenario. I then built and deployed a signed/notarized release candidate from merged `master` on macOS 27.0 build `26A5378j` and ran the new harness with distribution trust enforcement enabled.

Automated evidence passed:
- release doctor and identity/signing contracts
- Apple notarization submission `8bea2f66-2959-4fcf-b28d-a8f034794697` accepted
- stapling and Gatekeeper (`Notarized Developer ID`)
- `keypath-cli system inspect --json`: `planStatus=ready`, `driverCompatible=true`, `isOperational=true`, `userActionRequired=false`
- VirtualHID extension `[activated enabled]`
- KeyPath-owned Kanata launchd job running
- KeyPath and Kanata processes present
- TCP readiness on `127.0.0.1:37001`

[...]

Issue #747, opened 2026-06-06
Verify permission and setup UX on clean accounts and signed releases


malpern:
## Context

KeyPath depends on macOS permissions and service-management state that can be confusing during first run and upgrade: Accessibility, Input Monitoring, Full Disk Access, Login Items, privileged helper registration, and the Kanata launch daemon. The final signed build is healthy on the current machine, but clean-account testing must verify that the app explains permission blockers clearly.

## Scope

This is a focused signed-build validation pass, not a redesign task. If testing exposes confusing or incorrect behavior, split that behavior into a concrete issue rather than implementing broad wizard polish here.

## Test plan

- [ ] Validate first-run setup on a clean account where permissions are not already granted.
- [ ] Validate upgrade behavior where permissions are already granted.
- [ ] Confirm Input Monitoring denial is detected and explained correctly.
- [ ] Confirm Accessibility denial is detected and explained correctly.
- [ ] After granting Accessibility or Input Monitoring, confirm KeyPath recognizes the grant without a relaunch when macOS permits it.
- [ ] If macOS requires a relaunch before a new grant becomes visible, confirm the wizard offers an explicit **Restart KeyPath** action instead of polling indefinitely.
- [ ] Confirm Login Items/pending approval is represented distinctly from runtime failure.
- [ ] Confirm Fix It does not report success unless runtime is healthy or approval is explicitly pending.
- [ ] Confirm the status screen gives a concrete next action when macOS blocks setup.

## Acceptance criteria

- [ ] Permission-related blockers are understandable to a non-developer tester.
- [ ] The app does not conflate macOS approval-pending states with internal failures.
- [ ] Permission polling either observes the grant or transitions to accurate restart guidance.
- [ ] Any confusing or incorrect UX found here gets split into a focused issue.

Issue #740, opened 2026-06-06
Post-1.0: continue targeted error taxonomy cleanup


malpern:
## Context

Split out from #731 after the stable-1.0 release blockers were fixed.

The broad audit called out many local error enums. #731 corrected that not every local enum is bad: some are domain-specific and should stay local.

## Direction

Do targeted cleanup only where it improves behavior, observability, caller contracts, or user-visible diagnostics.

## Candidate Areas

- overlapping config/runtime errors,
- places where `KeyPathError` was intended to replace older app-wide errors,
- reload/save/apply failures that should preserve the applied/pending/rejected/failed contract from #732.

## Acceptance Criteria

- [ ] Inventory overlapping error enums in config/runtime/apply paths.
- [ ] Identify specific consolidations that improve caller behavior or diagnostics.
- [ ] Keep domain-specific local errors when they carry useful context.
- [ ] Add or update tests where caller-visible error semantics change.

## Non-goals

- Do not collapse every local error enum into one global type.
- Do not change user-facing error messages incidentally.

Issue #172, opened 2026-02-06
Upstream: Virtual HID keyboard doesn't emit native media key usage pages (F-row)


malpern:
## Summary

When Kanata intercepts keyboard input on macOS, the F-row keys (F1-F12) lose their default media functions (brightness, volume, playback, Mission Control, Spotlight, etc.). This happens because the Karabiner DriverKit virtual keyboard sends raw HID function key codes rather than the Apple-specific media key usage pages that macOS expects.

## Upstream Issue

[jtroo/kanata#975 — F-row keys get overridden on macOS](<https://github.com/jtroo/kanata/issues/975>)

## Root Cause

The `psych3r/driverkit` virtual HID keyboard device does not replicate the physical Apple keyboard's behavior of sending media key HID usage pages (Consumer Control page 0x0C) for F-row keys. Instead, it only sends standard Keyboard page (0x07) function key codes. macOS treats these as raw F1-F12 regardless of the user's "Use F1, F2, etc. as standard function keys" system preference.

The fix would need to happen in the **driverkit virtual keyboard driver** — making it natively support Consumer Control usage page reports so the virtual keyboard can emit brightness, volume, and media transport events the way a physical Apple keyboard does.

## KeyPath Status: Solved at App Layer

KeyPath already works around this with a built-in **"macOS Function Keys" rule collection** (`RuleCollectionCatalog.swift`) that:

* Maps F1-F12 → `brdn`, `brup`, `prev`, `pp`, `next`, `mute`, `vold`, `volu` in media mode
* Uses `push-msg` + `ActionDispatcher` for system actions Kanata can't natively emit (Mission Control, Spotlight, Dictation, DND)
* Provides a UI toggle between media keys and standard F-keys (`FunctionKeysView.swift`)
* Is enabled by default so users get working media keys out of the box

**This workaround is not upstreamable to Kanata** — it relies on KeyPath's ActionDispatcher for system actions. The real fix requires driver-level changes in `psych3r/driverkit`.

## Why Track This

* If driverkit adds native Consumer Control page support, KeyPath could simplify its F-row handling (remove push-msg workarounds for Mission Control/Spotlight/Dictation)
* The `fn` key toggle behavior (switching between media and F-keys) also doesn't work through the virtual keyboard — a driver fix could restore this
* macOS Sequoia added fn-based window management shortcuts that don't work through the virtual HID at all

## Upstream Status

[...]

PR #1168, opened 2026-07-13
Honor backlog status labels in issue dashboard

malpern:
## Summary
- derive issue-dashboard execution state from `on-hold` and `tracking-only` labels
- remove stale hard-coded issue-number priority overrides
- add focused precedence tests for held and tracking issues

## Why
The backlog now uses explicit GitHub labels as its source of truth. The dashboard still classified held feature ideas as queued, feature, or human work because it ignored those labels and carried obsolete issue-number exceptions.

## Validation
- `python3 Scripts/lab/tests/issue-dashboard-tests.py`
- generated a fresh state file: 1 queued, 2 human-led, 74 deferred
- `git diff --check`
- review gate: remote review gate selected; GitHub `claude-review` required

### DIFF
--- a/Scripts/lab/tests/issue-dashboard-tests.py
+++ b/Scripts/lab/tests/issue-dashboard-tests.py
@@ -19,17 +19,17 @@ def issue(number: int, *labels: str) -> dict:
 
 
 class IssueDashboardTests(unittest.TestCase):
-    def test_active_and_next_override_general_labels(self) -> None:
-        self.assertEqual(module.issue_status(issue(748, "tech-debt")), "active")
-        self.assertEqual(module.issue_status(issue(848, "testing", "agent-ok")), "next")
-
     def test_upstream_release_wait_is_human_gated(self) -> None:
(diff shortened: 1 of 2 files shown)


malpern:
Confirmed and documented the precedence intentionally in b0a2590f:

- `on-hold` / `tracking-only` are explicit execution-state labels and therefore outrank work type and operator.
- `wwdc26` is now only a topic label; it does not imply human gating.
- Active #919 remains `human` through its explicit `human-in-loop` label, while held WWDC feature ideas remain `deferred` through `on-hold`.

Added direct regression coverage for both WWDC cases. All 10 dashboard tests pass.

malpern:
Verified before merge:

- #172 is open with both `human-in-loop` and `on-hold`.
- #747 and #919 are open with `human-in-loop`.
- #740 is closed, so it is no longer part of the open dashboard.
- The current automation state contains zero `health` or `status` values equal to `waiting`.
- PR #1168 only changes the two issue-dashboard files; the waiting-state removal noted in review came from a newer base commit outside this diff.

Rebased onto current `origin/master` (branch is now 0 behind).

PR #1182, opened 2026-07-15
Fix wizard service action false failures

malpern:
## Summary
- retain Start, Restart, and Stop results in the runtime wizard page
- re-observe live runtime state after every action while suppressing only issues captured before a successful action
- refresh canonical wizard state after both successful and failed actions
- add behavioral regression coverage for stale-issue precedence and fresh process verification

## Root cause
The service page discarded the lifecycle coordinator result, refreshed live runtime state, and then evaluated it against stale wizard issues captured before the action. A successful restart could therefore render as failed even after runtime readiness and input capture had been verified.

## Validation
- `TEST_FILTER=ServiceStatusEvaluatorActionTests ./Scripts/run-tests-safe.sh`
- `TEST_FILTER=WizardKanataServicePageLogParsingTests ./Scripts/run-tests-safe.sh`
- `./Scripts/test-fast.sh --changed` (596 XCTest cases passed; an unrelated Swift Testing shared-state failure passed in isolation)
- `TEST_FILTER=KarabinerConflictService ./Scripts/run-tests-safe.sh`
- `./Scripts/review-gate.sh` (remote review gate selected)


malpern:
Addressed the review findings in 601817fe:

- successful actions now re-query current runtime status instead of force-setting the requested state, so a Boolean/status mismatch still fails
- stale pre-action issues are suppressed only when the action succeeded; a fresh process observation is always required
- failed actions now also refresh parent wizard state
- removed the duplicate start/restart test and brittle source-grepping test
- added behavioral tests for stale-issue suppression, fresh-process mismatch, and failed-action diagnostics

malpern:
Follow-up in c1ee803f narrows successful-action suppression to pre-action daemon issues only. Permission issues are retained and now have an explicit regression test.

`onRefresh()` launches a fresh parent system inspection and does not assign this page's local `serviceStatus`; it cannot overwrite the immediate post-action observation. Keeping it is intentional so the canonical wizard snapshot replaces the stale issues and can route a healthy install to Summary.

malpern:
Addressed the provenance concern in 4e830a9d. Start, Restart, and Stop now capture `issuesBeforeAction` synchronously before awaiting the lifecycle operation. Post-action suppression can therefore affect only daemon evidence known before that action; the fresh canonical inspection launched afterward owns any newly detected daemon issue.

malpern:
Simplified the final model in 7f217a00 based on the review discussion:

- a successful Start/Restart result already means `ServiceLifecycleCoordinator` verified readiness/input capture (or pending approval); the immediately-following fresh runtime observation distinguishes ready from pending/stopped
- a successful Stop still requires the fresh runtime observation to be stopped
- therefore pre-action wizard issues cannot override a successful lifecycle postcondition; they are intentionally excluded only for that immediate evaluation
- failed actions use the issue list at completion for diagnosis
- the trailing canonical parent inspection remains the source of all newly detected issues

Tests now explicitly cover stale daemon issues, stale permission issues, fresh runtime mismatch, and failed-action diagnostics.

malpern:
Confirmed from the implementation: this is not a launchctl/process-only success. `ServiceLifecycleCoordinator.verifyRuntimeReadinessAfterStart` builds a `KanataServiceRuntimeSnapshot` and returns `.ready` only when `snapshot.readiness.isReady`; the logged evidence explicitly includes running, TCP response, and `inputCaptureReady`. `startKanata` returns false if that readiness deadline fails. Restart delegates to that same start path.

The one true-without-readiness case is SMAppService pending approval; the immediately-following `currentRuntimeStatus()` is not `.running`, so `evaluateServiceCompletion` reports that the action did not complete rather than healthy. Stop likewise must match a fresh stopped observation. Thus a pre-action Kanata permission/capture issue contradicting a verified ready result is stale by the coordinator contract, not a currently-valid issue being masked.

PR #1185, opened 2026-07-16
Fix macOS 27 lab console session delivery

malpern:
## Summary
- add an owned `console-login` lab action for disposable macOS 27 Parallels leases
- verify the encrypted guest credential, catch `sysadminctl` false success, and require a real `keypathqa` console session after reboot
- add an RFB pointer-delivery probe that distinguishes protocol acknowledgement from delivered input
- document the remaining Apple Remote Management control prerequisite without treating it as a KeyPath failure

## Root cause
The base intentionally boots at `loginwindow`, while Parallels RFB acknowledges pre-login input without applying it. Two additional defects blocked autonomous recovery: the no-newline secret stream caused `read` under `set -e` to exit at EOF, and macOS 27 build `26A5378j` returned zero from `sysadminctl` while leaving automatic login disabled.

The new flow accepts neither transport acknowledgement nor a zero process exit as proof. It verifies account authentication, automatic-login state, `/dev/console` ownership, and cursor movement independently.

## Live validation
On an owned disposable macOS 27.0 (`26A5378j`) Parallels lease:
- `keypathqa` credential authentication passed
- the `sysadminctl` false-success path was detected
- the disposable-clone fallback established `keypathqa` as `/dev/console` owner after restart
- ARD/RFB authentication passed
- the pointer probe correctly failed because macOS reported event posting disabled and the cursor did not move
- artifacts were collected and the lease was destroyed

Apple's supported command-line Remote Management setup is view-only. Full RFB control remains gated on Remote Management approval through System Settings or MDM; ordinary Screen Sharing toggling was tested and was insufficient on this build.

## Validation
- `/bin/bash -n Scripts/lab/keypath-lab`
- `git diff --check`
- branch behind count: `0`
- review gate: `remote review gate selected`
- GitHub `claude-review`: passed
- code quality and accessibility checks: passed
- cold CI path completed in 19m07s; the only failure was the unrelated load-sensitive `SystemValidatorTests` wall-clock assertion also observed on #1186
- unchanged warm rerun: full `build-and-test` passed in 2m47s

### DIFF
--- a/Scripts/lab/keypath-lab
+++ b/Scripts/lab/keypath-lab
@@ -14,6 +14,8 @@ Commands:
   preflight
   create --macos 15|26|27 --lane managed-functional|unmanaged-ui --commit SHA --installer PATH [--ttl 2h] [--desktop]
   install-app LEASE_ID
+  console-login LEASE_ID
+  rfb-pointer-probe LEASE_ID --x X --y Y
   desktop-bootstrap LEASE_ID [--install-tools]
   nameplate LEASE_ID enable|show|hide|status
   protected-click LEASE_ID --app APP --window TITLE (--x X --y Y | --ax-x X --ax-y Y) [--after-window TITLE]
@@ -125,6 +127,28 @@ EOF
         [[ $# -eq 1 ]] || usage
         require_lease_id "$1"
         remote install-app "$1"
+        ;;
+    console-login)
+        [[ $# -eq 1 ]] || usage
+        require_lease_id "$1"
+        remote console-login "$1"
+        ;;
+    rfb-pointer-probe)
+        lease=${1:-}
+        [[ -n $lease ]] || usage
+        require_lease_id "$lease"
+        shift
+        x=
+        y=
+        while [[ $# -gt 0 ]]; do
+            case "$1" in
+                --x) x=${2:-}; shift 2 ;;
+                --y) y=${2:-}; shift 2 ;;
+                *) usage ;;
+            esac
+        done
+        [[ $x =~ ^[0-9]+$ && $y =~ ^[0-9]+$ ]] || usage
+        remote rfb-pointer-probe "$lease" "$x" "$y"
         ;;
     desktop-bootstrap)
         lease=${1:-}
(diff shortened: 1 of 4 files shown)


malpern:
Addressed in 3274ce8e. The comments, tests, and lab documentation now scope the guarantee to controller arguments/logs and explicitly document that macOS's noninteractive dscl/sysadminctl commands briefly receive the credential in argv inside the disposable guest. The runbook also says not to enable process-argument capture during console-login and to destroy the clone afterward.

malpern:
Addressed in 8c5c1f21. The guest now opens the FIFO read/write before a bounded zsh read, so opening the FIFO cannot block outside the timeout. A failed SSH stream immediately cancels the controller-side guest-control process and records credential-stream-failed separately from credential mismatch. The shell suite includes the stream-failure regression, and a real local FIFO probe confirmed the no-writer read exits after the bound.

malpern:
Added the focused reliability coverage in fc9990c8: guest-control readiness timeout and post-reboot console-user failure now have explicit regression tests, while the live-validated sysadminctl false-success fallback has static contract assertions for the kcpassword branch and accepted macOS 27 status wording. I left the low-priority IP-resolution refactor and generalized cursor behavior out to keep this PR scoped to the delivery bug.

malpern:
Addressed both findings in d330682f. The controller now runs the SSH writer under a Perl alarm using the same configured bound as the guest FIFO read; a hanging writer regression proves the action returns within the bound. Secret detection now records credential-leak-detected, redacts the log, and emits a dedicated error, with a regression test confirming the disclosed fixture is not retained.

malpern:
Confirmed no trap collision. remote.sh dispatches exactly one action per process, and console-login is its own terminal dispatch branch. The only other controller EXIT trap is release_admission_lock inside create(); that path clears its trap before create returns and cannot be followed by console_login in the same invocation. protected_click already uses the same secure-temp trap lifecycle. No code change is needed for this concern.

malpern:
Final CI status: code-quality, docs classification, and claude-review pass. The required build-and-test job was automatically cancelled twice by ci.yml's 10-minute job timeout while the Mini was actively compiling Swift modules; no test failure was reported. The prior head fc9990c8 passed the full build-and-test lane. The final d330682f delta is limited to the controller SSH timeout, distinct credential-leak status, docs, and their passing shell regressions. I am not mixing a workflow-timeout change into this focused PR.

PR #1187, opened 2026-07-16
Unblock macOS 27 Remote Management input

malpern:
## Summary

- frame the encrypted QA credential as a complete console-login record
- add a secret-safe, paced Parallels key-event path for focused authorization sheets
- require an independent RFB cursor postcondition and record the proven macOS 27 capability in the lab dashboard

## Root cause

The decrypted dotenv value has no trailing newline, so the guest FIFO reader could wait indefinitely for a complete record. Separately, Parallels drops characters when a full credential is sent as one unpaced key-event batch. Sending one event at a time with a short delay reliably completes the supported Remote Management authorization flow.

## Validation

- live disposable macOS 27 build `26A5378j`: Remote Management enabled through System Settings and RFB cursor delivery independently observed
- `python3 Scripts/lab/tests/update-progress-dashboard-tests.py`
- `python3 Scripts/lab/tests/issue-dashboard-tests.py`
- `./Scripts/test-fast.sh --changed` (full safe fallback: 4,746 tests passed)
- `./Scripts/review-gate.sh` (remote review gate selected; GitHub `claude-review` required)

Related to #919; this does not complete the full macOS 27 regression lane.

### DIFF
--- a/Scripts/lab/keypath-lab
+++ b/Scripts/lab/keypath-lab
@@ -15,6 +15,7 @@ Commands:
   create --macos 15|26|27 --lane managed-functional|unmanaged-ui --commit SHA --installer PATH [--ttl 2h] [--desktop]
   install-app LEASE_ID
   console-login LEASE_ID
+  secure-console-submit LEASE_ID
   rfb-pointer-probe LEASE_ID --x X --y Y
   desktop-bootstrap LEASE_ID [--install-tools]
   nameplate LEASE_ID enable|show|hide|status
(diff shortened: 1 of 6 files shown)


malpern:
Addressed the review concern with explicit regression coverage. `Scripts/lab/remote.sh` already runs under `set -euo pipefail`, so a pre-output Python validation failure propagates through the pipeline and prevents Return from being sent. The new test supplies an unsupported uppercase credential, requires a nonzero result, and verifies that no Parallels key event—including Return—is emitted.

Issue #870, opened 2026-06-10
[1.1] Define rule dependency graph data model and builder


malpern:
Part of [#865](https://github.com/malpern/KeyPath/issues/865) — the dependency-awareness epic. **Start here**; the other child issues depend on this vocabulary.

## Goal

Define a pure-data RuleDependencyGraph value type that records typed capabilities, requirements, active providers, and potential providers without performing mutation or presentation.

The graph is rebuilt deterministically from current configuration. It is not persisted and is not a second mutable source of truth.

## Location and boundaries

- Put the pure graph, capability, requirement, contribution, and evidence types in **KeyPathRulesCore**.
- Keep catalog merging, user-facing copy, confirmed mutations, and dialogs in **KeyPathAppKit**.
- Use the existing UUID collection identity for 1.1. Do not introduce a repository-wide ID migration in this issue.

## Capability vocabulary

Start with the minimum typed distinctions required by current behavior:

- **layerContent(layer)** — a layer contains useful behavior.
- **layerActivation(layer)** — a functional path reaches or activates that layer.
- **keyAlias(.hyper)** — the configured Hyper alias exists.

Do not use hard-coded rule-to-rule dependencies or a single untyped string namespace. For example, HRL Toggles requires layer content named fun; it does not directly depend on the built-in Function collection.

Normalize open-ended layer identifiers by trimming whitespace and lowercasing at graph construction.

## Requirements and evidence

Keep capability identity coarse. Attach detail such as affected keys or configuration fields to a separate Requirement/Evidence value.

Example conceptually:

- capability: layerContent(fun)
- evidence: keys [a, ;]

This lets UI explain “2 keys use Function” without turning every key into a separate capability.

## Graph semantics

The graph must distinguish:

1. **Known providers** — all current/catalog collections capable of providing a capability, including disabled collections.
2. **Active providers** — known providers whose collections are enabled in the analyzed state.

This distinction is required for a later “Enable required rules” action: a missing capability has no active provider, but may have one or more disabled providers available to enable.

A clean split for sequencing is:

- #870 defines RuleDependencyContribution and builds a graph from contributions.
- #867 derives contributions from concrete RuleCollection families and adds the collection-facing adapter.

The graph should retain:

- enabled collection IDs;
- provides by collection;
- requirements by collection;
- reverse providers by capability;
- structured requirement evidence.

## Acceptance

- [ ] Types live in KeyPathRulesCore.
[...]

Issue #867, opened 2026-06-10
[1.1] Encode per-family dependencies (HRL Toggles, Launcher, Leader-based, Custom)


malpern:
Part of [#865](https://github.com/malpern/KeyPath/issues/865). Depends on the capability and graph model in #870.

## Goal

Encode explicit per-family dependency contributions. This is product logic that inspects each RuleCollection and its current configuration and emits:

- capabilities it can provide;
- capabilities it currently requires;
- structured evidence for each requirement.

Keep this logic explicit and testable. Do not hide it behind reflection or a generic string registry.

## Provider semantics

| Source | Provides |
|---|---|
| A collection containing useful mappings on a layer | layerContent(targetLayer) |
| A functional base → layer activator, including Vim Navigation or configured Leader Key | layerActivation(targetLayer) |
| A functional chained activator such as nav → fun | layerActivation(targetLayer) |
| Caps Lock Remap with hold output exactly equal to Hyper | keyAlias(.hyper) |

Built-in collections are not exclusive providers. A custom collection or future pack may satisfy the same capability.

Contributions are derived for known enabled and disabled collections. The graph decides whether a provider is active.

## Requirement semantics

| Family/configuration | Requires |
|---|---|
| Home Row Layer Toggles | One layerContent capability per distinct enabled-key assignment |
| Home Row Mods in layer mode | One layerContent capability per distinct enabled-key assignment |
| Quick Launcher in holdHyper mode | keyAlias(.hyper) |
| A collection activated from a non-base source layer | layerActivation(sourceLayer) |
| A collection whose mappings live directly on a non-base layer without its own activator | layerActivation(targetLayer) |
| Leader-sequence activation | The corresponding activation capability; keep sequence-prefix granularity coarse for 1.1 |
| Custom mappings on a non-base layer | layerActivation(targetLayer) |
| All other configurations | No prerequisites |

Both HRL “While Held” and “Toggle” modes require useful target-layer content. Parser acceptance or stub layer declaration is not evidence that the target layer does useful work.

## Evidence

Requirements should carry structured context when available:

- HRL/Home Row Mods: affected enabled keys grouped by target layer.
- Activator chains: source and target layers.
- Quick Launcher: activation mode and Hyper trigger setting.
- Custom mappings: affected mapping IDs or keys.

Do not generate user-facing sentences here.

## Acceptance

- [ ] Add a collection-facing contribution extractor and graph-builder adapter using #870’s pure types.
[...]

PR #1194, opened 2026-07-24
Add forward and reverse prerequisite detection

### LINKED ISSUE #866: [1.1] Implement Prerequisite detector: forward + reverse queries on RuleCollectionsManager
Part of [#865](https://github.com/malpern/KeyPath/issues/865). Depends on #870 and the concrete contribution extraction in #867.

## Goal

Add RuleCollectionsManager+PrerequisiteDetection.swift beside the existing conflict detector. It performs two single-hop analyses against a shared RuleDependencyGraph:

- **Forward:** what would be missing after enabling or saving this proposed collection/configuration?
- **Reverse:** what newly becomes unsatisfied after disabling this provider?

Analysis describes consequences only. It does not mutate collections or format dialog copy.

## Proposed API shape

The exact names can be refined in the PR, but results must remain structured.

### Prerequisite

- consumer collection ID;
- missing Capability;
- structured Requirement/Evidence;
- all available provider collection IDs;
- any unambiguous recommended provider.

### DependentMapping

- dependent collection ID;
- newly unsatisfied Capability;
- structured Requirement/Evidence;
- remaining available providers, if any.

Do not put userFacingDescription strings in these domain results. Presentation belongs to the dialog adapter.

## Proposed-state semantics

Queries must analyze the state the user is proposing, before mutation:

- enabling a currently disabled candidate;
- replacing an existing collection’s configuration with an edited candidate;
- disabling/removing one provider while preserving all other providers.

A reverse warning is emitted only when the proposed removal leaves no active provider for a requirement. If another active provider satisfies the same capability, there is no orphan.

A forward warning is emitted for newly introduced missing requirements. Unrelated edits must not repeatedly warn about a pre-existing missing requirement.

## Determinism

Return results in stable order by consumer display order/name, capability, and evidence. Never expose dictionary/set iteration order to tests or UI.

## Acceptance

- [ ] Module follows the neighboring conflict-detection organization and naming conventions.
- [ ] Forward analysis uses the proposed candidate configuration, not stale manager state.
- [ ] Reverse analysis removes only the proposed victim’s contribution before evaluating consumers.
- [ ] Known disabled providers are returned as available fixes.
- [ ] Multiple providers are handled correctly.
- [ ] No warning when another active provider still satisfies the capability.
- [ ] Results contain structured evidence and no user-facing prose.
- [ ] Unit tests cover: default catalog; missing layer content; missing activation path; missing Hyper alias; available disabled provider; multiple providers; no orphans; one orphan; multiple orphans; proposed config edit; stable ordering.

## Out of scope

- Mutation or auto-resolution.
- UI work.
- Automatic cascading disable.
- Transitive resolution.

malpern:
## Summary

- add structured `RulePrerequisite` and `RuleDependentMapping` domain results beside the existing collection conflict detector
- analyze forward enable/save proposals by replacing an existing candidate or appending a brand-new collection in a proposed graph snapshot
- analyze reverse disables by removing only the selected provider contribution
- suppress pre-existing missing capabilities and avoid warnings when another active provider remains
- return disabled alternatives, an unambiguous recommendation when exactly one provider exists, and deterministic result ordering

## Why

#870 defined the dependency graph and #867 connected real collection families to it. This PR adds the proposed-state queries needed by the later forward and reverse dialog work without adding presentation or mutation behavior.

## Impact

No UI or collection mutation changes. The new APIs describe dependency consequences only.

## Checks

- `TEST_FILTER=RuleCollectionsManagerPrerequisiteDetectionTests ./Scripts/run-tests-safe.sh` — 16 focused tests passed
- `./Scripts/test-fast.sh --changed` — 643 tests passed, 5 skipped
- `./Scripts/test-full.sh` — 4,792 tests passed
- `python3 Scripts/check-accessibility.py` — passed
- SwiftFormat 0.61.1 and `git diff --check` — passed
- local review gate selected the required remote review path; `claude-review` must pass before merge

Closes #866

### DIFF
--- a/Sources/KeyPathAppKit/Services/RuleCollections/RuleCollectionsManager+PrerequisiteDetection.swift
+++ b/Sources/KeyPathAppKit/Services/RuleCollections/RuleCollectionsManager+PrerequisiteDetection.swift
@@ -1,0 +1,165 @@
+import Foundation
+import KeyPathRulesCore
+
+/// A requirement that would be newly unsatisfied by a proposed collection.
+///
+/// This value is deliberately presentation-free. Dialogs and other clients
+/// decide how to describe the capability and its structured evidence.
+struct RulePrerequisite: Equatable, Sendable {
+    let consumerCollectionID: UUID
+    let missingCapability: RuleCapability
+    let requirement: RuleRequirement
+    let availableProviderCollectionIDs: [UUID]
+    let recommendedProviderCollectionID: UUID?
+}
+
+/// A dependent requirement that would become unsatisfied after removing a provider.
+struct RuleDependentMapping: Equatable, Sendable {
+    let dependentCollectionID: UUID
+    let newlyUnsatisfiedCapability: RuleCapability
+    let requirement: RuleRequirement
+    let remainingAvailableProviderCollectionIDs: [UUID]
+}
+
+extension RuleCollectionsManager {
+    /// Finds requirements newly introduced by enabling or saving `proposedCollection`.
+    ///
+    /// An existing collection is replaced in place so its display order remains
+    /// stable. A new collection is appended. The proposal is always analyzed as
+    /// enabled because this query runs before the enable/save mutation.
+    func prerequisites(
+        for proposedCollection: RuleCollection
+    ) -> [RulePrerequisite] {
+        let adapter = RuleCollectionDependencyGraphAdapter()
+        let currentGraph = adapter.build(
+            collections: ruleCollections,
+            customRules: customRules
+        )
+
+        var candidate = proposedCollection
+        candidate.isEnabled = true
+
+        var proposedCollections = ruleCollections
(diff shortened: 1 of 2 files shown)


malpern:
The transitive-chain concern is valid future work, but it is intentionally outside this PR's contract. #866 explicitly defines these as **single-hop analyses** and lists both **automatic cascading disable** and **transitive resolution** as out of scope; #870 likewise names transitive resolution/SAT solving as an explicit non-goal. The current implementation therefore evaluates whether another enabled provider directly satisfies each structured capability, matching the issue's proposed-state semantics.

If KeyPath introduces a real A → B → C rule chain, that should get a separate graph-viability design ticket rather than silently changing this foundational query's semantics here. The newly requested brand-new-candidate branch is now covered by `testForwardAnalysisAppendsBrandNewCandidate`; all final checks are green.

PR #1219, opened 2026-07-25
Make key repeat speed direction predictable

### LINKED ISSUE #1213: [Timing] Make Key Repeat speed direction and units predictable
Part of #1203 and #888.

## User problem

The Key Repeat control is labeled Speed but edits the interval in milliseconds directly. Moving the slider toward a larger value makes repetition slower, so the control's direction contradicts its label.

## Approved vocabulary

- **Repeat start delay**: how long a key is held before automatic repeat begins.
- **Repeat speed**: repeats per second, shown as the primary value.
- Millisecond interval remains an expert/secondary detail.

## Scope

- Bind the speed control so moving right/higher always means faster repetition.
- Present repeats per second as the primary value and interval milliseconds secondarily.
- Preserve precise round-trip behavior and define rounding for supported values.
- Rename Delay to Repeat start delay and add concise felt-effect copy.

## Acceptance criteria

- [ ] Moving the speed control right increases repeats per second.
- [ ] The primary displayed and accessibility value is repeats/sec.
- [ ] The equivalent millisecond interval is available as secondary detail.
- [ ] Minimum/maximum labels describe slow-to-fast direction correctly.
- [ ] Conversion, clamping, rounding, persistence, and generated output have focused tests.
- [ ] Key Repeat user documentation is updated.

## Non-goals

- Changing Kanata's repeat implementation.
- Adding acceleration curves or per-key repeat profiles.

malpern:
Closes #1213

## What changed
- present repeat speed in repeats/sec while preserving Kanata millisecond intervals
- make moving the speed slider right consistently faster
- distinguish repeat start delay from repeat speed in labels, summaries, and accessibility values
- document the unit conversion and update preset examples
- cover conversion, clamping, rounding, persistence, and generated Kanata output

## Validation
- KeyRepeatSpeedScaleTests: 5 passed
- KanataConfigurationGeneratorSnapshotTests: 17 passed
- accessibility identifier scan: passed across 356 files
- live UI: verified labels, values, direction, 33 to 50 repeats/sec mapping, and restored original 30 ms setting
- full safe suite: 4,839 passed; one unrelated SystemValidator timing assertion missed its 250 ms budget under full snapshot load, then passed alone in 20 ms
- review gate: remote review selected; do not merge until claude-review passes

### DIFF
--- a/Sources/KeyPathAppKit/UI/Rules/KeyRepeatControlView.swift
+++ b/Sources/KeyPathAppKit/UI/Rules/KeyRepeatControlView.swift
@@ -196,6 +196,13 @@ struct KeyRepeatControlView: View {
 
             ScrollView {
                 VStack(alignment: .leading, spacing: 20) {
+                    Text(
+                        "Repeat start delay is how long you hold a key before it begins repeating. Repeat speed is how quickly it continues.",
+                        bundle: #bundle
+                    )
+                    .font(.callout)
+                    .foregroundStyle(.secondary)
+                    .fixedSize(horizontal: false, vertical: true)
                     allKeysSection
                     arrowSection
                     HStack(alignment: .top, spacing: 12) {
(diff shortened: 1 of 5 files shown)


malpern:
Verified both review questions against the full tree:

1. The updated Careful row corrects stale documentation. The shipped `.careful` config already uses 300 ms / 25 ms for all four arrow overrides (`RuleCollectionConfiguration.swift`, lines 958-966); this PR does not change those preset values.
2. A repo-wide search at exact head `a005a6268dd05abe3a881171e9c99d287d76a7f5` finds no remaining `keysPerSecond` references, and CI build/test is green. The public helper had no external-package consumer in this repository.

No follow-up code change is needed.

PR #1233, opened 2026-07-28
Route CLI service control through helper

malpern:
## What changed

- route CLI Kanata start and stop operations through narrowly scoped privileged-helper methods
- bump the helper compatibility contract to 1.2.0
- bootstrap lifecycle dependencies and verify transitions with fresh runtime evidence
- account for the LaunchDaemon throttle window during startup verification
- add focused lifecycle coverage and a lint preventing future direct privileged CLI execution

## Why

`keypath-cli service stop`, `start`, and `restart` invoked `launchctl` directly. Normal users therefore received an authorization failure even though the bundled CLI is an authorized client of KeyPath's helper. Live deployment also exposed false-negative postconditions caused by missing CLI bootstrap state and cached pre-transition snapshots.

## Impact

Normal-user CLI service control now uses KeyPath's existing privilege boundary without `sudo` or an administrator password. Lifecycle commands report success only after observing the requested runtime state.

## Validation

- `TEST_FILTER=CLIServiceTests TIMEOUT_SECONDS=180 ./Scripts/run-tests-safe.sh`
- `TEST_FILTER=CLIPrivilegeBoundaryLintTests TIMEOUT_SECONDS=120 ./Scripts/run-tests-safe.sh`
- installed `/Applications/KeyPath.app/Contents/MacOS/keypath-cli service restart --json` returned `{"restarted": true}`
- final installed state: helper 1.2.0 fresh, Kanata running, VirtualHID healthy, app signature valid

PR #1234, opened 2026-07-28
Handle KeepAlive service restart race

malpern:
## Summary
- make CLI restart verify the final healthy runtime instead of requiring a transient stopped snapshot
- continue requiring both privileged helper stop and start calls to succeed
- add regression coverage for launchd KeepAlive relaunching before the CLI observes stopped state

## Why
Installed verification after #1233 reproduced a false failure: `service restart --json` returned code 6 while launchd had already relaunched Kanata and the service became fully operational moments later. The SMAppService job is KeepAlive, so an intermediate stopped snapshot is not a reliable restart postcondition.

## Validation
- `TEST_FILTER=CLIServiceTests TIMEOUT_SECONDS=180 ./Scripts/run-tests-safe.sh` (10 XCTest cases, 0 failures; runner reports 11 passed)
- live reproduction on the installed merged #1233 build
- final installed status recovered to `isOperational: true`, helper 1.2.0 fresh, Kanata/VHID healthy

### DIFF
--- a/Sources/KeyPathAppKit/CLI/SystemFacade.swift
+++ b/Sources/KeyPathAppKit/CLI/SystemFacade.swift
@@ -98,11 +98,27 @@ public struct SystemFacade: Sendable {
     }
 
     public func restartService() async -> Bool {
-        guard await stopService() else { return false }
-        if restartDelayNanoseconds > 0 {
-            try? await Task.sleep(nanoseconds: restartDelayNanoseconds)
+        await MainActor.run {
+            CLIRuntimeBootstrap.ensureConfigured()
         }
-        return await startService()
+        do {
+            try await stopServiceOperation()
+            await runtimeCacheInvalidator()
+            if restartDelayNanoseconds > 0 {
+                try? await Task.sleep(nanoseconds: restartDelayNanoseconds)
+            }
+            try await startServiceOperation()
+            await runtimeCacheInvalidator()
+
+            // The SMAppService job is KeepAlive. launchd may relaunch it before
+            // polling can observe a stopped snapshot, so restart success is the
+            // final healthy runtime—not a transient stopped state.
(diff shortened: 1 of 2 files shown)


malpern:
Review follow-up:

- Added the missing negative-path coverage: if the helper-backed start operation throws after a successful stop request, restart reports failure. The focused suite now executes 11 CLI service tests with 0 failures. The existing test already covers the stop-operation throw path.
- A normally returning stop operation means the helper received a zero exit status from `launchctl kill`. Start then targets the same `system/com.keypath.kanata` launchd job. launchd owns and serializes that single job instance, so this does not create a second unmanaged Kanata process.
- We intentionally do not require a stopped runtime snapshot during restart because the service is KeepAlive and may be relaunched before the poll observes the transient gap; that exact behavior caused the installed false failure this PR fixes.

malpern:
Second review follow-up:

This is an intentional lifecycle invariant change, not a dropped verification step. Both helper operations address the same launchd-owned job (`system/com.keypath.kanata`); `startKanataService` executes `launchctl kickstart` for that job and cannot create an independently managed second instance. Requiring a sampled stopped state is incompatible with a KeepAlive job because launchd may legitimately replace the process before the 200 ms poll sees the gap.

The mutating restart remains postcondition-verified: after both helper calls return success, cache state is invalidated again and the CLI waits up to 20 seconds for the final `running && responding` runtime. A failed helper reply throws through `executeXPCCall`, and the CLI returns its existing structured restart failure. The generic `Bool` facade already intentionally maps start/stop errors to CLI error code 6; preserving detailed helper errors would be a separate API-shape improvement across all three lifecycle commands.

On failed start, the cache has already been invalidated after the successful stop, so downstream status cannot reuse pre-stop state. There is no new mutation after the failed helper reply to invalidate.

PR #1235, opened 2026-07-28
Make helper service lifecycle idempotent

malpern:
## Summary
- add one privileged-helper restart operation backed by `launchctl kickstart -k`
- route CLI restart through that atomic launchd operation and verify the final healthy runtime
- make helper stop idempotent when the registered job has no current process
- bump the helper contract and bundle version to 1.3.0

## Why
Installed acceptance of #1234 captured the concrete remaining race: immediately after fast deploy, Kanata was registered but in launchd `spawn scheduled` state. The helper stop returned `No process to signal`, so CLI restart failed even though launchd later restored a healthy runtime. A native `kickstart -k` restart avoids the stop/start transition window entirely.

## Validation
- `TEST_FILTER=CLIServiceTests TIMEOUT_SECONDS=180 ./Scripts/run-tests-safe.sh` — 10 XCTest cases, 0 failures
- `TEST_FILTER=HelperVersionContractTests TIMEOUT_SECONDS=120 ./Scripts/run-tests-safe.sh` — 2 cases, 0 failures
- `TEST_FILTER=CLIPrivilegeBoundaryLintTests TIMEOUT_SECONDS=120 ./Scripts/run-tests-safe.sh` — 1 case, 0 failures
- reproduced helper error from unified log: `stopKanataService failed ... No process to signal` while `launchctl print` showed the job registered and subsequently running

### DIFF
--- a/Sources/KeyPathAppKit/CLI/SystemFacade.swift
+++ b/Sources/KeyPathAppKit/CLI/SystemFacade.swift
@@ -12,11 +12,11 @@ public struct SystemFacade: Sendable {
 
     private let startServiceOperation: @Sendable () async throws -> Void
     private let stopServiceOperation: @Sendable () async throws -> Void
+    private let restartServiceOperation: @Sendable () async throws -> Void
     private let runtimeCacheInvalidator: @Sendable () async -> Void
     private let runtimeSnapshotProvider: @Sendable () async -> RuntimeSnapshot
     private let runtimeTransitionTimeoutSeconds: TimeInterval
     private let pollDelayNanoseconds: UInt64
-    private let restartDelayNanoseconds: UInt64
 
     public init() {
         self.init(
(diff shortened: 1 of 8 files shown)


malpern:
Review follow-up:

1. Restart intentionally operates on an already registered service. The prior helper-backed **start** operation also used `launchctl kickstart` and did not register/bootstrap the SMAppService job, so this PR does not remove a bootstrap fallback. Registration/repair remains the installer path; a restart of an unregistered job should fail and route the user to repair rather than silently broaden a lifecycle command into installation.
2. Agreed that the helper subprocess boundary is not mocked by `CLIServiceTests`. The helper is deliberately tested here through its version/protocol contract and bounded subprocess implementation; after CI, acceptance requires installing helper 1.3.0 and running the real CLI restart against the live system job. I will not merge/deploy-complete without that live result.
3. The matcher is based on the exact observed helper failure: `stopKanataService failed: ... No process to signal.` `BoundedProcess` assigns the same pipe to stdout and stderr and returns their combined text as `result.out`, so the guard sees this message regardless of which stream launchctl used.

PR #1253, opened 2026-07-30
Fix CLI service lifecycle through SMAppService

malpern:
## Summary
- route CLI service start, stop, and restart through the owning SMAppService lifecycle
- verify unregister removes both the registration and process, with a bounded retry and helper-backed stale-job cleanup
- bump the 1.0.1 release candidate to build 12 and document the live acceptance finding

## Validation
- 16 focused lifecycle/CLI tests passed
- Developer ID signed build 12 passed the identity contract and strict codesign verification
- live build 11 to build 12 replacement refreshed the registered runtime to build 12
- one stop remained down across five observations over 10 seconds
- one start returned operational with build 12, helper 1.3.2, and VirtualHID 8 healthy

## Release scope
This is the final lifecycle correction required before the 1.0.1 Sparkle release candidate is rebuilt and notarized. Physical HID proof remains a separate lab gate and is not claimed here.

### DIFF
--- a/Sources/KeyPathApp/Info.plist
+++ b/Sources/KeyPathApp/Info.plist
@@ -11,7 +11,7 @@
 	<key>CFBundleDisplayName</key>
 	<string>KeyPath</string>
 	<key>CFBundleVersion</key>
-	<string>11</string>
+	<string>12</string>
 	<key>CFBundleShortVersionString</key>
 	<string>1.0.1</string>
 	<key>CFBundlePackageType</key>
--- a/Sources/KeyPathAppKit/CLI/SystemFacade.swift
+++ b/Sources/KeyPathAppKit/CLI/SystemFacade.swift
@@ -21,13 +21,13 @@ public struct SystemFacade: Sendable {
     public init() {
         self.init(
             startServiceOperation: {
-                try await HelperManager.shared.startKanataService()
+                try await KanataDaemonService.shared.start()
             },
             stopServiceOperation: {
-                try await HelperManager.shared.stopKanataService()
+                try await KanataDaemonService.shared.stop()
             },
             restartServiceOperation: {
-                try await HelperManager.shared.restartKanataService()
+                try await KanataDaemonService.shared.restart()
             },
             runtimeCacheInvalidator: {
                 await MainActor.run {
--- a/Sources/KeyPathAppKit/CLI/SystemFacade.swift
+++ b/Sources/KeyPathAppKit/CLI/SystemFacade.swift
@@ -110,8 +110,8 @@ public struct SystemFacade: Sendable {
             try await restartServiceOperation()
             await runtimeCacheInvalidator()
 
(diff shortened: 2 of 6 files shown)


malpern:
Addressed both lifecycle review findings in 69a0c2aea:

- Stop now performs one fresh SMAppService status read per bounded cleanup phase and polls only launchd/process evidence between phases, rather than issuing fresh status IPC on every 100 ms iteration.
- Start now distinguishes pending approval, failed registration persistence, and failed process/TCP readiness; it cannot report success until registration is enabled and runtime readiness is proven. Restart inherits the same verified start contract.
- Added explicit approval-required and registered-but-not-ready tests. The focused CLI/lifecycle suite is now 18/18 green, and pinned SwiftFormat passes.

malpern:
Addressed the follow-up stop-state concern in 88f470aa4. The postcondition now distinguishes registration still present from a removed registration with a lingering process. Privileged cleanup is followed by another owning-API unregister only when registration remains; the final gate still requires both registration removal and process exit, so a merely killed process cannot be reported as a durable stop. The tests now exercise the real registration/process branches (no composite postcondition bypass), including stale registration and process-only lingering cases. Focused suite: 19/19 green; pinned formatting: green.

malpern:
Correction: the follow-up fix commit is 928dae5bb (the behavior and 19/19 validation described above are unchanged).

malpern:
Addressed the shutdown-grace concern in 76a85e101: ordinary SMAppService stop now waits six seconds, slightly longer than the launchd plist's five-second exit timeout, before invoking privileged cleanup. Start retains its 20-second readiness window. The real lingering-process fallback test now exercises that full grace period; focused suite remains 19/19 green and pinned formatting passes.

malpern:
Addressed the transient post-fallback unregister concern in ec081c062: that cleanup call is now best-effort, and the final registration+process postcondition remains authoritative. Removed the test-only unregister-error suppression and added a test where the third unregister throws after cleanup but the real final state is healthy. Focused suite: 20/20 green; pinned formatting: green.

malpern:
Addressed the initial-unregister concern and completed the same state-authoritative contract for all lifecycle mutations in 888372d12: initial unregister, register, and privileged cleanup delivery errors no longer preempt a fresh final-state check. Added coverage for initial unregister failure, register-throws-after-enabling, and privileged-cleanup error with successful final recovery. Focused suite: 23/23 green; pinned formatting: green.

malpern:
Final integration confirmation:

- Callers await the lifecycle operation; `SystemFacade` then independently verifies the final runtime state. The bounded worst-case retry window does not violate a caller timeout or assume an immediate kickstart.
- CLI start/stop/restart failures are converted to actionable `serviceControlFailed` guidance directing users to fresh status and the repair UI. No external exhaustive matches depend on the new internal `KanataDaemonServiceError` cases.

All required checks are green; proceeding to merge and exact-merge release acceptance.

PR #1254, opened 2026-07-30
Fix setup wizard runtime spinner and prepare 1.0.2 hotfix

malpern:
## What changed

- Recheck runtime status once per second only while the service page is in a transient startup state.
- Stop polling immediately when the runtime settles.
- Cap transient checks at 30 and surface an actionable Restart state instead of an endless spinner.
- Remove the obsolete disabled timer scaffolding.
- Bump the hotfix to KeyPath 1.0.2 build 13.

## Root cause

The runtime service page performed a single status observation. Its broad auto-refresh timer had previously been disabled to avoid invasive permission checks. If the one observation landed during the brief runtime startup window, the page rendered “Starting KeyPath runtime…” forever even after the runtime became healthy.

## User impact

Users who launch KeyPath during that transition can now progress automatically once the already-healthy runtime is observed. The retry is bounded and runtime-only; it does not restore broad permission polling.

## Validation

- Reproduced on the notarized 1.0.1 build while the CLI and validator showed a healthy runtime.
- Relaunching the UI confirmed the runtime remained continuously operational and the main interface appeared.
- `swift test --filter ServiceStatusEvaluatorActionTests` — 7 passed.
- Pinned SwiftFormat 0.61.1 lint — passed.
- `plutil -lint Sources/KeyPathApp/Info.plist` — passed.
- `git diff --check` — passed.

### DIFF
--- a/Sources/KeyPathApp/Info.plist
+++ b/Sources/KeyPathApp/Info.plist
@@ -11,9 +11,9 @@
 	<key>CFBundleDisplayName</key>
 	<string>KeyPath</string>
 	<key>CFBundleVersion</key>
-	<string>12</string>
+	<string>13</string>
 	<key>CFBundleShortVersionString</key>
-	<string>1.0.1</string>
+	<string>1.0.2</string>
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleIconFile</key>
--- a/Sources/KeyPathInstallationWizard/Core/ServiceStatusEvaluator.swift
+++ b/Sources/KeyPathInstallationWizard/Core/ServiceStatusEvaluator.swift
@@ -26,14 +26,35 @@ public enum ServiceStatusEvaluator {
         completedAttempts: Int
     ) -> Bool {
         guard completedAttempts < transientRefreshAttemptLimit else { return false }
+        return isTransientStartupStatus(
+            runtimeStatus: runtimeStatus,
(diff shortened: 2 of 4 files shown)


malpern:
Addressed in `45e0d3c55`: retry and exhaustion now share one transient-state predicate, covering both `.starting` and `.stopped` inside the startup window. Added a regression assertion proving transient `.stopped` reaches the actionable timeout state, then reran the focused suite (7/7) and pinned formatting.

malpern:
Follow-up on the latest review notes:

1. Confirmed the retry path is runtime-only and does not invoke permission checks. `RuntimeCoordinator.currentRuntimeStatus()` delegates to `ServiceLifecycleCoordinator.currentRuntimeStatus()`, which calls `KanataDaemonService.refreshStatus()` (cached `SMAppService.status` + PID evidence). `isInTransientRuntimeStartupWindow()` uses timestamps and cached SMAppService management state. Neither calls `PermissionOracle` or a TCC mutation path.
2. Overlapping loops are prevented: `WizardKanataServicePage.refreshStatus()` begins with `refreshTask?.cancel()` before assigning the replacement task; `onDisappear` also cancels it.
3. Exhaustion is now a shared pure decision (`didExhaustTransientStatus`) used directly by the page, with regression coverage for the previously missed transient `.stopped` case.

No additional permission-polling surface is being restored.

PR #1258, opened 2026-08-08
Support secure dialog input on the Parallels macOS 26/27 lanes

malpern:
Granting Accessibility or Input Monitoring on macOS 27 raises an authenticated password prompt. `secure-dialog-input` refused every lane but Tart macOS 15, so **the denied-to-allowed transition could not be automated at all on the lanes KeyPath and Typover actually test** — including every "newly granted" row in `docs/testing/macos-27-regression.md`.

## Why the Tart path could not be reused

It requires `peekaboo` and `mcporter` under `/opt/homebrew` **inside the guest**, and the Parallels desktop base has no Homebrew at all.

So this delivers the credential with the key-event transport `secure_console_submit` already proves out, which needs nothing installed in the guest. The plaintext is still read only from an owner-only temp file and never enters argv, logs, the guest pasteboard, or an artifact, and the accepted alphabet stays exactly as narrow.

## Delivery is not success

The sheet-closed postcondition is preserved, matching the Tart lane — without it, a credential that never reached the field reports a pass. It runs through the peekaboo the base installs at a stable path.

**Verified that the assertion bites:** defeating the postcondition makes the suite fail with `authentication sheet did not close`, and the command reports `passed` when it should not.

## Live verification on macOS 27

Not just unit-tested. On a real lease, driving Typover's first-run permission flow:

```
secure_dialog_input      passed
credential_transport     parallels-key-events
```

Then confirmed independently rather than trusting the exit status:

- the Accessibility toggle for Typover switched **on** in System Settings
- after relaunch Typover reports `Accessibility, Allowed`
- Input Monitoring still reports `Not yet allowed`, so the grant was specific rather than a blanket change

## Note for whoever automates these flows next

`peekaboo click` did **not** reliably invoke Typover's SwiftUI button — it timed out while the sheet stayed closed. `peekaboo perform-action --action AXPress` worked first try. That matches what `p01-unmanaged-input-monitoring-spike.md` already recorded for KeyPath's wizard controls, so it is now two projects with the same finding: prefer `perform-action` over `click` for SwiftUI targets.

### DIFF
--- a/Scripts/lab/remote.sh
+++ b/Scripts/lab/remote.sh
@@ -882,12 +882,110 @@ run_command() {
   return "$exit_code"
 }
 
+# Run peekaboo inside a Parallels guest as the console user. The base installs
+# it at a stable path because the guest has no Homebrew, and AX queries only see
+# the session when they run as the console user rather than as root.
+parallels_guest_peekaboo() {
+  local resource=$1 parallels_cli=$2
+  shift 2
+  local guest_peekaboo=${KEYPATH_LAB_GUEST_PEEKABOO:-/usr/local/bin/peekaboo}
+  local quoted
+  printf -v quoted '%q ' "$@"
+  # prlctl exec mangles the arguments of the first command in the string, so
+  # lead with a no-op before the command that matters.
+  "$parallels_cli" exec "$resource" /bin/zsh -lc \
+    "true; sudo -u keypathqa $guest_peekaboo $quoted"
+}
+
+# macOS 26 and 27 run on Parallels, whose guests carry neither the peekaboo nor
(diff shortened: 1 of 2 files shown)


malpern:
Superseded by #1259.

The Parallels macOS 26/27 secure dialog input from this PR was ported into the vm-lab repository (commit `f0ca542`) along with its tests, and the sheet-closed postcondition was re-verified to fail when defeated. #1259 deletes the `Scripts/lab/remote.sh` this PR modifies, so merging this afterwards would resurrect a deleted file.

No work is lost — the behaviour now lives where the lab does.

PR #1292, opened 2026-09-06
Clarify current Karabiner migration boundaries

malpern:
## What changed

Clarifies the current Karabiner migration boundary without promising future parity. The migration guide now distinguishes supported, limited, and deferred Karabiner capabilities, removes the temporary-loss and unsupported 95% conversion language, and documents importer-specific limits plus skipped-rule review and original-config backup guidance. The installation guide links to the boundary section.

## Validation

- `git diff --check`
- Focused frontmatter/link/scope assertions
- Preserved-feature source checks
- `PATH=/usr/bin:/bin ./Scripts/review-gate.sh` → remote review gate selected (exit 2)

No Swift build, tests, publishing, or merge performed.

### DIFF
--- a/guides/installation-wizard.md
+++ b/guides/installation-wizard.md
@@ -63,7 +63,7 @@ Similar to Accessibility — macOS asks you to grant Input Monitoring access.
 
 If you have Karabiner-Elements installed, the wizard offers to import your existing rules. You can review which rules will convert and choose which to import.
 
-See [Switching from Karabiner]({{ '/migration/karabiner-users/' | relative_url }}) for details on what converts.
+See [Switching from Karabiner]({{ '/migration/karabiner-users/' | relative_url }}) for the current conversion and support boundaries.
 
 ### 5. Start service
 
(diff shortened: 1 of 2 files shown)


malpern:
Confirmed against the current implementation. `keypath import karabiner` declares its positional input as `karabiner.json or complex_modifications rule file` in `Sources/KeyPathCLI/Commands/Import/ImportKarabinerCommand.swift:13`.

The normal path is `CollectionsFacade.importFromKarabiner` (`Sources/KeyPathAppKit/CLI/CollectionsFacade.swift:226-231`), which calls `KarabinerConverterService.convert`. That service decodes `KarabinerConfig` with `JSONDecoder` (`Sources/KeyPathAppKit/Services/Karabiner/KarabinerConverterService.swift:105-112`) and converts a selected profile (`:79-100`). If that fails, the CLI-only fallback accepts a JSON object containing `rules`, wraps it as a selected Karabiner profile, and passes it back through the same converter (`CollectionsFacade.swift:285-307`).

Goku EDN is therefore outside this importer path: the accepted inputs are JSON, and there is no EDN parser or Goku conversion branch in the command/facade/converter path. Skipped rules are carried into the CLI result at `CollectionsFacade.swift:265-273`; the CLI prints each description/reason at `ImportKarabinerCommand.swift:110-115`, and the wizard surfaces the skipped count at `WizardKarabinerImportPage.swift:139-143`. No inline review comments were present.

malpern:
Rechecked after rebase at f83dc4ce: the earlier implementation evidence still applies. ImportKarabinerCommand.swift lines 110–115 iterates result.skippedRules and prints each description and reason, so this is per-rule reporting, not only a count. The accepted JSON/full-profile and complex-modifications fallback paths, and absence of an EDN parser on that path, are detailed in the preceding source confirmation. Both behavioral claims are verified.

PR #1300, opened 2026-09-15
Fix two service-lifecycle defects found during the physical-HID proof

malpern:
Two defects surfaced while running the P02 physical-HID proof on a disposable
macOS 26 guest. Both leave the user believing something worked when it did not.

## A failed restart left the daemon unregistered

`keypath service restart` could leave `com.keypath.kanata` absent from the
launchd system domain entirely, with no command-line path back. On the guest,
`launchctl` then reported no such service and `keypath system repair` returned
`userActionRequired`; only the app UI recovered it.

Cause: `restart()` is `stop()` then `start()`. Stop removes the SMAppService
registration *before* it verifies its postcondition, and start swallows a failed
re-register with `try?`. Either path can throw with the registration already
gone, and nothing rolls it back.

Restart now restores the registration before surfacing the failure, so a restart
that cannot succeed leaves the caller no worse off than before it ran. A
registration merely awaiting approval is left alone, since re-registering would
discard the pending approval.

Start also polls briefly while a fresh registration settles. SMAppService
registers asynchronously, so the status read taken immediately after `register()`
can still report `notRegistered` and fail a start that would have worked a moment
later. That is the likely cause of recovery needing two attempts after an
emergency stop, which is what we observed.

Also corrects two messages that actively misdirected this investigation: the
restart hint blamed administrator authorization without mentioning the service
state, and the runtime wizard said "Click Fix" while its buttons read Start and
Restart.

## A rule for a collection-owned key was persisted, then refused

With the built-in Home Row Arrows collection enabled, which claims `f` as a
momentary layer activator:

```bash
keypath rule add f --tap f --hold lsft --on-conflict replace --apply
```

reported the rule as created, and the apply that followed then refused. The rule
stayed in `CustomRules.json` and broke every later `keypath config apply` until
removed by hand. `--on-conflict replace` could not help, because `addRule`
compared the new rule only against other custom rules and never against
collections.

The conflict detector itself works correctly and fires as intended (#667); it
simply fires far too late, after the write. The facade now pre-flights the
prospective rule set through that same detector before anything is written, and
`rule add` renders the refusal as an exit-4 conflict naming the owning collection.
Conflicts already present between collections are subtracted, so a store in a bad
state does not block an unrelated key. The app-facing initializers supply no
collection loader, so GUI behavior is unchanged.

## Verification

25 tests across the two areas, including 3 new daemon regression tests covering
rollback after a failed stop, rollback after a failed register, and the settling
poll; and 9 new tests covering the conflict refusal, the skip strategy, and the
cases that must still succeed. Adjacent rule, config-generation, deduplication,
and lifecycle suites all pass.

Note: the repo's pinned test scripts could not run here, because the bundle at
`Xcode-26.6.app` currently contains Xcode 27 and the pin guard rejects it. Tests
were run with `swift test` directly. That drift is worth a separate look.


malpern:
CI note: `build-and-test` failed at the **Verify runner disk reserve** step with exit 75, before compiling anything.

```
disk_reserve_free_gib	66
disk_reserve_minimum_gib	90
disk_reserve_busy: free space is below the 90 GiB safety reserve
```

Exit 75 is this project's infrastructure-busy signal, so this is a wait on runner disk rather than a rejection of the change. `claude-review` and `code-quality` both pass.

The runner's own `actions-runner-cleanup.sh` was run and freed nothing: CI scratch already lives on the external volume, which has about 3 TB free. The shortage is on the mini's startup volume, from data unrelated to CI. Re-run once that has headroom.

